PR #175, opened 2014-05-24
Attempt implementing NaN-checks for floating-point variables.

ghost:
There's no need to merge this, I'm just testing things out and trying to figure out how much more complicated this would be to implement in a 3rd party tool rather than in the compiler as in [this pull request](https://github.com/D-Programming-Language/dmd/pull/3573).


MartinNowak:
> Don't get me even started on tabs. I don't know what it is with editors. ¯\_(ツ)_/¯

There is an editorconfig.

PR #218, opened 2015-01-06
Make xml more conformant w.r.t. empty tags and newlines

John-Colvin:
Empty tags shouldn't have anything between open and close, not even a newline.
If all opening tags (except the XML header tag) lead with a newline and all closing tags end with a newline, you get correct results.

A small tweak to the <attribute> tag printing was needed to avoid printing a newline after the attribute text but before the closing tag.

### DIFF
--- a/src/astprinter.d
+++ b/src/astprinter.d
@@ -12,7 +12,7 @@ import std.array;
 
 template tagAndAccept(string tagName)
 {
-    immutable tagAndAccept = `output.writeln("<` ~ tagName ~ `>");`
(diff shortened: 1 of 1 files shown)


John-Colvin:
Whoops, botched this completely, I'll ping when I've fixed it.

John-Colvin:
Ok, so it's not technically broken. It produces correct xml, but it's ugly because it also produces excess blank lines.

Hackerpilot:
Closing this because it's very out-of-date.

PR #221, opened 2015-01-17
avoid FileException to be thrown

BBasile:
if invalid path/file is passed in the args.


src/main.d
```diff
@@ -260,7 +260,9 @@ string[] expandArgs(string[] args)
 	if (args.length == 1)
 		args ~= ".";
 	foreach (arg; args[1 ..$])
-	{
+	{        
+		if (!exists(arg))
```
Hackerpilot: I'm not sure that failing silently is the best idea.

PR #205, opened 2015-01-23
git submodules don't work with dub

MartinNowak:
Submodules also aren't included in github's src packages.
I'd recommend making libdparse a separate dub package, which is also cleaner when only the lexer/parser libraries are needed in another project.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -5,8 +5,10 @@
 	"authors": ["Brian Schott"],
 	"license" : "Boost Software License - Version 1.0",
 	"targetType": "executable",
+	"preBuildCommands": ["git log -1 --format='%H' > githash.txt"],
+	"stringImportPaths": ["."],
 	"dependencies": {
-		"libdparse": { "path": "libdparse/" },
-		"inifiled": { "path": "inifiled/" }
-	}
+		"libdparse": "~>0.2.0",
+		"inifiled": ">=0.0.3",
+	},
 }


MartinNowak:
Would be great to release a libdparse 0.1.2 which could be then be used by dscanner.
Using branch dependencies (~master) breaks fairly often.

SerialVelocity:
This would be great to merge. I can't currently do:
```bash
$ dub fetch dscanner
$ dub run dscanner
```

SerialVelocity:
This probably doesn't work because it is missing something like
```json
"dflags": [
  "-J.",
],
```

SerialVelocity:
Actually, the git hash in the file doesn't exist so I guess it would also need a pre-build command too

MartinNowak:
Updated

MartinNowak:
By fully using dub you could replace [build.bat](https://github.com/Hackerpilot/Dscanner/blob/679c1b4e609fc3eeecc7946ab83731e75d09ee9d/build.bat), [dsc.sublime-project](https://github.com/Hackerpilot/Dscanner/blob/679c1b4e609fc3eeecc7946ab83731e75d09ee9d/dsc.sublime-project), [makefile](https://github.com/Hackerpilot/Dscanner/blob/679c1b4e609fc3eeecc7946ab83731e75d09ee9d/makefile#L25), and [test.sh](https://github.com/Hackerpilot/Dscanner/blob/679c1b4e609fc3eeecc7946ab83731e75d09ee9d/test.sh) and can get a nice CI intregration with minimal effort (#222).
It also becomes simpler for anyone else to use your project.

MartinNowak:
Thanks

MartinNowak:
Ah, stupid me. `git log` won't work with dub packages fetch from the registry because they aren't git repos. Any other ideas how to tag the version, maybe @s-ludwig?

SerialVelocity:
If it isn't a git repo, parse it from the folder name? I know it's a bit hacky but it fixes the issue for now.

MartinNowak:
Also though of that, but the folder name is `dscanner-master` currenty :(.

MartinNowak:
I think we need a proper approach for that, in some dub issue it was proposed, that dub always generates a version module which can be included from the current package.

PR #237, opened 2015-04-28
Generate etags

smolt:
Here is a starting point for generating Emacs style etags with -e.  Based on the existing ctags module.  I can imagine more, but this should get things going.  Can be used like this:

```
$ dscanner >TAGS -e `find ~/dmd2/src/{druntime,phobos} -name \*.d`
```
Works best in emacs with tags-case-fold-search set to nil so that things like `M - . File` finds struct File in std.stdio first instead of some struct member named "file" elsewhere. 

I use hard tabs in source because it looks like other files in repo were that way.  Hope that's right.

### DIFF
--- a/src/etags.d
+++ b/src/etags.d
@@ -1,0 +1,235 @@
+// Emacs tags based on ctags.d with this header:
+//			Copyright Brian Schott (Hackerpilot) 2012.
+// Distributed under the Boost Software License, Version 1.0.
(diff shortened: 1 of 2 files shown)


Hackerpilot:
The style settings for the project are in the .editorconfig file: https://github.com/Hackerpilot/Dscanner/blob/master/.editorconfig

MartinNowak:
Thanks a lot.

PR #246, opened 2015-05-08
updated .editorconfig with the bracket style used in Dscanner.

joakim-brannstrom:


.editorconfig
```diff
@@ -5,3 +5,6 @@ indent_size = 4
 tab_width = 4
 trim_trailing_whitespace = true
 indent_style = tab
+
+dfmt_brace_style=allman
+dfmt_space_after_cast=false
```
Hackerpilot: Have I been forgetting spaces after casts? (Not that this matters. dfmt doesn't do anything with this option yet.)

joakim-brannstrom: Not from what I have seen :)
I put it as true in my personal configuration. So when updating Dscanners I thought it would be best to put in the "default value" according to dfmt's specification (dfmt_space_after_cast 	true, false 	false).

PR #247, opened 2015-05-08
Support for exuberant ctags format used by vim.

joakim-brannstrom:
Basic support. Tested with for example vim plugin TagBar.
Gives a result that is good enough to be useful in every day work.

### Third value
According to the specification for Exuberant Ctags the third value can be either a line number or an ex mode command. I chose for simplicities sake the line number in this implementation.
In the future I would like to change the behaviour to be ex mode commands instead of line numbers.
Example:
Bar	stub.d	/^    struct Bar {$/;"	s	line:34	struct:Foo	access:public

Do you have any suggestion as to how to get everything from the start of the line up to and including the AST node?

### AccessState
I'm not really happy with the current solution of a state. I would like to, if possible, to implement it with token look-ahead.

### TemplateDeclaration
Current implementation generates tags for variables directly inside a template, which is wrong. But we want it to generate other types of tags in templates, nesting and all the fun. So... work to be done.

### DIFF
--- a/src/ctags.d
+++ b/src/ctags.d
@@ -13,11 +13,13 @@ import std.range;
 import std.stdio;
 import std.array;
 import std.conv;
+import std.typecons;
 
 /**
  * Prints CTAGS information to the given file.
+ * Includes metadata according to exuberant format used by Vim.
  * Params:
- *     outpt = the file that CTAGS info is written to
+ *     outpt = the file that Exuberant TAGS info is written to
  *     fileNames = tags will be generated from these files
  */
 void printCtags(File output, string[] fileNames)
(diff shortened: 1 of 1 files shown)


Hackerpilot:
I wrote the original ctags module based on the documentation on the exuberant ctags site. Is there any reason that this needs to exist in parallel instead of being a replacement or improvement for the current ctags code?

joakim-brannstrom:
Yes, you are right. The standard for "exuberant" is actually just an extension of ctags with more metadata. I've updated the pull request to extend your ctags implementation.

src/ctags.d
```diff
@@ -28,101 +30,143 @@ void printCtags(File output, string[] fileNames)
...
-			~ "!_TAG_PROGRAM_URL\thttps://github.com/Hackerpilot/Dscanner/\n");
+	output.write(
+		"!_TAG_FILE_FORMAT\t2\n" ~ "!_TAG_FILE_SORTED\t1\n" ~ "!_TAG_FILE_AUTHOR\tBrian Schott\n" ~ "!_TAG_PROGRAM_URL\thttps://github.com/Hackerpilot/Dscanner/\n");
 	tags.sort().copy(output.lockingTextWriter);
 }
 
 private:
 
-void doNothing(string, size_t, size_t, string, bool) {}
+/// States determining how an access modifier affects the parsning.
```
Hackerpilot: "parsing"

src/ctags.d
```diff
@@ -28,101 +30,143 @@ void printCtags(File output, string[] fileNames)
...
+/// Reset, when ascending the AST reset back to the previous access.
```
Hackerpilot: If you put these lines inside the enum body, ddoc will associate the comments with the enum values.

src/ctags.d
```diff
@@ -28,101 +30,143 @@ void printCtags(File output, string[] fileNames)
...
+}
+
+string paramsToString(Dec)(const Dec dec)
+{
+	import std.d.formatter : Formatter;
+
+	auto app = appender!string();
+	auto formatter = new Formatter!(typeof(app))(app);
+
+	static if (is(Dec == FunctionDeclaration))
```
Hackerpilot: You can write this as `is(Dec == FunctionDeclaration) || is(Dec == Constructor)` and then delete lines 82-85.

joakim-brannstrom:
Thanks for the feedback :+1: 
Done as you suggested.
Do you want me to squash the commits to keep the git history in Dscanner clean?

PR #248, opened 2015-05-09
Etags updates

smolt:
Change default --etags behavior so that private/package declarations are not tagged.  Added a new option --etagsAll to include private/package stuff.  Also create tags for modules and aliases (that was definitely needed) and some fixes too.

I was not sure if --etagsAll was best command line option or if there should have been another option.  Let me know.


Hackerpilot:
`--etagsAll` is fine with me.

smolt:
Thanks!

PR #309, opened 2016-03-11
Added a test for static if...else if mistakes.

dhasenan:
If you write your code as:
```
static if (foo)
  doFoo();
else if (bar)
  doBar();
```

This check will catch it. If you write it as:
```
static if (foo)
  doFoo();
else {
  if (bar)
    doBar();
}
```

This isn't a bug and we won't warn about it.

### DIFF
--- a/src/analysis/static_if_else.d
+++ b/src/analysis/static_if_else.d
@@ -1,0 +1,95 @@
+// Copyright Chris Wright (dhasenan) 2016.
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file LICENSE_1_0.txt or copy at
+//          http://www.boost.org/LICENSE_1_0.txt)
+
+module analysis.static_if_else;
+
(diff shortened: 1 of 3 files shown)


src/analysis/static_if_else.d
```diff
@@ -0,0 +1,95 @@
...
+		addErrorMessage(ifStmt.line, ifStmt.column, KEY, "Mismatched static if. Use 'else static if' here.");
+	}
+
+	const(IfStatement) getIfStatement(const ConditionalStatement cc)
+	{
+		if (cc.falseStatement.statement)
+		{
+			if (cc.falseStatement.statement.statementNoCaseNoDefault)
+			{
+				if (cc.falseStatement.statement.statementNoCaseNoDefault.ifStatement)
```
Hackerpilot: This is a good reminder that I need to make libdparse more user-friendly.

PR #311, opened 2016-03-14
Don't warn about renamed imports.

### LINKED ISSUE #310: Don't warn about renamed local imports
The local import check incorrectly warns about this:

```D
void foo() {
  import stdio = std.stdio;
}
```

The problem with a bare local import is that it can shadow an open-ended set of declarations. Selective imports fix that, and that's what Dscanner recommends. Renamed imports also fix that, but Dscanner doesn't realize that.

dhasenan:
Renamed imports can't shadow an arbitrarily large set of declarations, so don't warn about them.

Fixes #310.

### DIFF
--- a/src/analysis/local_imports.d
+++ b/src/analysis/local_imports.d
@@ -52,10 +52,16 @@ class LocalImportCheck : BaseAnalyzer
 		if ((!isStatic && interesting) && (id.importBindings is null
 				|| id.importBindings.importBinds.length == 0))
 		{
-			addErrorMessage(id.singleImports[0].identifierChain.identifiers[0].line,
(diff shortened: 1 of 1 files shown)


src/analysis/local_imports.d
```diff
@@ -75,3 +81,21 @@ private:
...
+	import analysis.config : StaticAnalysisConfig;
+
+	StaticAnalysisConfig sac;
+	sac.local_import_check = true;
+	assertAnalyzerWarnings(q{
+		void testLocalImport()
+		{
+			import std.stdio; // [warn]: Local imports should specify the symbols being imported to avoid hiding local symbols.
+			import std.fish : scales, head;
+			import DAGRON = std.experimental.dragon;
```
Hackerpilot: Strong Mad? You just keep doing your thing, man.

PR #319, opened 2016-04-14
add --enforce flag to error on warnings

wilzbach:
I imagine this to be useful if a project uses DScanner for linting and wants to `enforce` their style guide.


wilzbach:
Closed as this will is expected to be the default behavior.

PR #340, opened 2016-05-07
Added an option that alows to skip the unittests analysis

BBasile:
I find that a lot of irrelevant warnings are generated because of the particular nature of _unittests_ (e.g variable not used, variable not modified so could be const, etc).   
This PR modifies the BaseAnalyzer so that the visit is optionally skipped within a _Unittest_.

The feature can be activated with `--skipTests` or in the configuration file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,12 +43,28 @@ analysis and it does not compile the code.
 The "--styleCheck" or "-S" option runs some basic static analysis checks against
 the given source files.
 
+#### Skip style checks in the tests
+Static checks in the unit tests can produce irrelevant warnings. For example,
+it's legit to declare a variable that's not used if the goal is to verify that
+a templatized function can be instantiated by inference of the type of this variable.
+To avoid these cases, it's possible to pass the "--skipTests" option.
+
 #### Configuration
 By default all checks are enabled. Individual checks can be enabled or disabled
 by using a configuration file. Running ```dscanner --defaultConfig``` will
 generate a default configuration file and print the file's location. The
 "--config" option will allow you to specify the path to a configuration file if
 you do not want to use the one created by the "--defaultConfig" option.
+
+For each check, three values are possible:
+* `"disabled"`: the check is not performed.
+* `"enabled"`: the check is performed.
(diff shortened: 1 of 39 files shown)


BBasile:
I don't know why Travis-CI fails. By default the option is not activated so it does not change the current behaviour.

BBasile:
I don't know If I really need to argue for this PR but for example, 
`dscanner experimental/logger/multilogger.d -S` outputs

> experimental/logger/multilogger.d(107:12)[warn]: Local imports should specify the symbols being imported to avoid hiding local symbols.
> /home/basile/Dev/repos/D-Programming-language/phobos/std/experimental/logger/multilogger.d(115:10)[warn]: Variable n0_1 is never modified and could have been declared const or immutable.
> experimental/logger/multilogger.d(120:10)[warn]: Variable n1_1 is never modified and could have been declared const or immutable.
> experimental/logger/multilogger.d(134:36)[warn]: Variable line is never modified and could have been declared const or immutable.
> experimental/logger/multilogger.d(163:27)[warn]: Variable line1 is never modified and could have been declared const or immutable.
> experimental/logger/multilogger.d(169:26)[warn]: Variable line2 is never modified and could have been declared const or immutable.

`dscanner experimental/logger/multilogger.d -S --skipTests` outputs
>  (nothing)


Clear enough.

Hackerpilot:
It might be better to change the behavior of the configuration file so that a user can specify
```
enum_array_literal_check="skip-unittest"
number_style_check="disabled"
builtin_property_names_check="enabled"
```

BBasile:
I don't know IniFile at all. At first sight it looks like the properties must be changed to enum, which may break the reloading of any configuration file (is it OK ?).

BBasile:
Done.

BBasile:
PR rebased.

BBasile:
PR rebased.

BBasile:
I've realized (only) today that the old ini file could be reloaded without problem, simply by renaming "enabled" and "disabled" (defined in an enum) to "true" and "false". 

Does anyone wants a fixup ?

PR #344, opened 2016-05-14
add allman style check

wilzbach:
checks for the Phobos allman style

https://en.wikipedia.org/wiki/Indent_style#Allman_style

I couldn't find a way how to do this without looking at the raw code buffer.


wilzbach:
Superseded by #348 - sorry about the noise.

PR #348, opened 2016-05-15
Whitespace + allman checks

### LINKED ISSUE #324: check for more than two consecutive empty lines
It should be possible to check whether there are two or more lines.

```
auto a  = 1;

auto b = 2;
```

or:

``
auto a()
{

}

auto b
{

}
```

This is also an guideline from Phobos.

### LINKED ISSUE #328: add check for trailing whitespace
Like it is done in dmd & Phobos:

See also:

https://github.com/D-Programming-Language/dmd/blob/master/src/checkwhitespace.d

### PR DESCRIPTION
**Pull request by wilzbach**
Adds the ability to check for whitespace between operators (#318);

I realized it makes sense to bundle my changes in one PR. Thus it includes #344, #345, #346 and the whitespace checks

Fixes:
- #345 (trailing whitespace) fixes #328 
- #346 (consecutive lines) fixes #324

### DIFF
--- a/src/analysis/allman.d
+++ b/src/analysis/allman.d
@@ -1,0 +1,221 @@
+// Distributed under the Boost Software License, Version 1.0.
+//	  (See accompanying file LICENSE_1_0.txt or copy at
+//			http://www.boost.org/LICENSE_1_0.txt)
+
+module analysis.allman;
+
+import dparse.lexer;
+import dparse.ast;
+import analysis.base : BaseAnalyzer;
+import dsymbol.scope_ : Scope;
+
+/**
+Checks for the allman style (braces should be on their own line)
+
+------------
+if (param < 0) {
+
+}
+------------
+
+should be
+
+------------
+if (param < 0)
+{
+
+}
+------------
+*/
+class AllManCheck : BaseAnalyzer
+{
+	///
+	this(string fileName, const(ubyte)[] code, bool skipTests = false)
+	{
+		super(fileName, null, skipTests);
+		this.code = code;
+	}
+
+	override void visit(const WhileStatement st)
+	{
+		if (st.declarationOrStatement !is null)
+			checkForBrace(st.declarationOrStatement, st.expression.line, st.expression.column);
+	}
+
+	override void visit(const ForeachStatement st)
+	{
+		checkForBrace(st.declarationOrStatement, st.low.line, st.low.column);
+	}
+
+	override void visit(const ForStatement st)
+	{
+		checkForBrace(st.declarationOrStatement, st.test.line, st.test.column);
+	}
+
+	override void visit(const DoStatement st)
+	{
+		// the DoStatement only knows about the line and column of the expression
+		checkForBrace(st.statementNoCaseNoDefault, 0, 0);
+		st.statementNoCaseNoDefault.accept(this);
+	}
+
+	override void visit(const IfStatement st)
+	{
+		checkForBrace(st.thenStatement, st.expression.line, st.expression.column);
+		if (st.elseStatement !is null)
+			checkForBrace(st.elseStatement, st.expression.line, st.expression.column);
+	}
+
+	alias visit = ASTVisitor.visit;
+
+private:
+
+	const(ubyte)[] code;
+
+	enum string KEY = "dscanner.style.allman";
+	enum string MESSAGE = "Braces should be on their own line";
+
+	void checkForBrace(const DeclarationOrStatement declOrSt, size_t line, size_t column)
+	{
+		if(auto stst = declOrSt.statement)
+		{
+			checkForBrace(stst.statementNoCaseNoDefault, line, column);
+		}
+		declOrSt.accept(this);
+	}
+
+	void checkForBrace(const StatementNoCaseNoDefault st, size_t line, size_t column)
+	{
+		if(st !is null)
+		{
+			findBraceOrNewLine(st.startLocation, st.endLocation, line, column);
+		}
+	}
+
+	/**
+	Checks whether a brace or newline comes first
+	*/
+	void findBraceOrNewLine(size_t start, size_t end, size_t line, size_t column)
+	{
+		import std.algorithm : canFind;
+		import std.utf : byCodeUnit;
+
+		auto codeRange = (cast(char[]) code[start..end]).byCodeUnit;
+
+		// inline statements are allowed -> search for newline
+		if (codeRange.canFind('\n'))
+		{
+			foreach (s; codeRange)
+			{
+				// first brace
+				if (s == '{')
+				{
+					// DoStatement hasn't a proper line and column attached
+					// -> calculate ourselves
+					if (line == 0 && column == 0)
+					{
+						// find line & column of brace
+						auto t = findLineAndColumnForPos(start);
+						line = t.line + 1; // Dscanner starts lines at 1
+						column = t.column;
+					}
+					addErrorMessage(line, column, KEY, MESSAGE);
+					break;
+				}
+				// newline - test passed
+				else if (s == '\n')
+				{
+					break;
+				}
+			}
+		}
+	}
+
+	/**
+	Counts all matches of an symbol and the number of iterated characters
+	*/
+	auto findLineAndColumnForPos(size_t pos)
+	{
+		import std.utf : byCodeUnit;
+		import std.typecons : tuple;
+
+		auto textBefore = (cast(char[]) code[0..pos]).byCodeUnit;
+		size_t line = 0;
+		size_t column = 0;
+
+		foreach (s; textBefore)
+		{
+			if (s == '\n')
+			{
+				line++;
+				column = 0;
+			}
+			else if (s != '\r')
+			{
+				// ignore carriage return
+				column++;
+			}
+		}
+		return tuple!("line", "column")(line, column);
+	}
+
+}
+
+unittest
+{
+	import analysis.config : StaticAnalysisConfig, Check;
+	import analysis.helpers : assertAnalyzerWarnings;
+	import std.format : format;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/analysis/allman.d [hunk 1]
```diff
@@ -0,0 +1,203 @@
...
+
+}
+------------
+*/
+class AllManCheck : BaseAnalyzer
+{
+    ubyte[] code;
+
+    ///
+    this(string fileName, ubyte[] code, bool skipTests = false)
```
Hackerpilot: What was the reason for taking the raw source code as a parameter here instead of the token array? The tokens are avalable in the analysis.run.analize function so they could be passed to this constructor.

wilzbach: I didn't know how I could access the tokens based on a statement and look ahead/before. I only found `startLocation`

src/analysis/allman.d
```diff
@@ -0,0 +1,203 @@
...
+                else if (s == '\n')
+                {
+                    break;
+                }
+            }
+        }
+    }
+
+    enum string KEY = "dscanner.style.allman";
+    enum string MESSAGE = "Brace should be on their own line";
```
Hackerpilot: Change to "its own line" or "Braces should" for singular/plural consistency.

src/analysis/consecutive_empty_lines.d
```diff
@@ -0,0 +1,90 @@
...
+
+    ///
+	this(string fileName, ubyte[] code, bool skipTests = false)
+	{
+		super(fileName, null, skipTests);
+		this.code = code;
+    }
+
+	override void visit(const Module)
+    {
```
Hackerpilot: Is there a mixture of spaces and tabs in this file?

wilzbach: Oh that's weird - sorry.
Will reformat it to spaces only.

src/analysis/space_between_operators.d
```diff
@@ -0,0 +1,343 @@
...
+
+    override void visit(const RelExpression expr)
+    {
+        checkForWhitespaceSymbols!(['<', '>', '=', '!'])(expr.line, expr.column - 1);
+        ASTVisitor.visit(expr);
+    }
+
+    mixin(WhiteSpaceMixin!("isWhitespace(expr.startIndex - 1); ASTVisitor.visit(expr);",
+                                "WhileStatement", "IfStatement", "ForStatement",
+                                "ForeachStatement"));
```
wilzbach: should I add functions and unittest blocks to this list?

src/analysis/space_between_operators.d
```diff
@@ -0,0 +1,343 @@
...
+            foreach (s;[1]) {}
+            foreach_reverse(s;[1]) {} // [warn]: Space between operators required
+            foreach_reverse (s;[1]) {}
+        }
+        void testExpression()
+        {
+            auto a = true ? 1:0; // [warn]: Space between operators required
+            auto a = true?1 : 0; // [warn]: Space between operators required
+        }
+	}c, sac);
```
Hackerpilot: There aren't any tests for unary operators. For example:
```d
auto a = -b;
a = ~mask & value;
b = *ptr;
```

src/analysis/consecutive_empty_lines.d
Hackerpilot: http://editorconfig.org/#download
https://github.com/Hackerpilot/Dscanner/blob/master/.editorconfig

wilzbach:
I managed to find some to rebase this :)

> What was the reason for taking the raw source code as a parameter here instead of the token array? The tokens are avalable in the analysis.run.analize function so they could be passed to this constructor.

The lexer [drops](https://github.com/Hackerpilot/libdparse/blob/master/src/dparse/lexer.d#L581) all unnecessary whitespace are and these checks explicitly depend on `lexWhitespace`.

(I excluded the more complex check for space between operators.)

Hackerpilot:
I don't recall if I asked this before, but wouldn't it be easier to check the tokens array for line numbers and column numbers? The tokens do have that information.

src/analysis/allman.d
```diff
@@ -0,0 +1,221 @@
...
+
+	override void visit(const ForStatement st)
+	{
+		checkForBrace(st.declarationOrStatement, st.test.line, st.test.column);
+	}
+
+	override void visit(const DoStatement st)
+	{
+		// the DoStatement only knows about the line and column of the expression
+		checkForBrace(st.statementNoCaseNoDefault, 0, 0);
```
wilzbach: For 

```d
do {

} while(true)
```

the AST is:

```
<doStatement>
  <expression>
    <unaryExpression>
      <primaryExpression>
        <true/>
      </primaryExpression>
    </unaryExpression>
  </expression>
  <statementNoCaseNoDefault>
    <blockStatement>
      <declarationsAndStatements>
        <declarationOrStatement>
          <statement>
            <statementNoCaseNoDefault>
              <expressionStatement>
                <expression>
                  <unaryExpression>
                    <primaryExpression>
                      <identifierOrTemplateInstance>
                        <identifier>i</identifier>
                      </identifierOrTemplateInstance>
                    </primaryExpression>
                    <suffix>++</suffix>
                  </unaryExpression>
                </expression>
              </expressionStatement>
            </statementNoCaseNoDefault>
          </statement>
        </declarationOrStatement>
      </declarationsAndStatements>
    </blockStatement>
  </statementNoCaseNoDefault>
</doStatement>
```

and none of the elements contains information about the line and column:

DoStatement: https://github.com/Hackerpilot/libdparse/blob/master/src/dparse/ast.d#L1467
BlockStatement: https://github.com/Hackerpilot/libdparse/blob/master/src/dparse/ast.d#L959
StatementNoCaseNoDefault: https://github.com/Hackerpilot/libdparse/blob/master/src/dparse/ast.d#L2249

wilzbach:
> I don't recall if I asked this before, but wouldn't it be easier to check the tokens array for line numbers and column numbers? The tokens do have that information.

Yes you did, but afaict the tokens array doesn't contain information about whitespace / new lines as they are already ignored during the lexing phase. However, I am happy if you prove me wrong ;-)

BBasile:
Try this, you must change the lexer config so that whites are included.

```D
void main()
{
    import dparse.lexer;
    import std.stdio, std.file : read;

    ubyte[] src = cast(ubyte[]) read(__FILE_FULL_PATH__);
    StringCache cache = StringCache(StringCache.defaultBucketCount);
    LexerConfig config = LexerConfig("", StringBehavior.source, WhitespaceBehavior.include, CommentBehavior.intern);

    const(Token)[] toks = getTokensForParser(src, config, &cache);

    foreach(tk; toks)
    {
        if (tk.type != tok!"identifier")
            writeln(str(tk.type));
        else if (tk.type != tok!"whitespace")
            writeln("whites for ", tk.text.length, " chars");
        else // much more case exist...
            writeln("'", tk.text, "'");
    }
}
```

but actually you only need them to detect the trailing spaces. Otherwise you can use the tokenN position, the tokenN length and the tokenN+1 position to detect missing whites after `if`, `foreach` and such tokens.
Also take care of the comments, since `if/*bla bla*/(true)` is perfectly valid.

BBasile:
I'd like a global switch to enable phobos-specific checks. A bit like the one for skipping the unittests. The problem is that when there's no configuration file, these phobos-specific-checks work. By default the phobos style is faschista. Dscanner is not made for phobos, even if it can be used to.

wilzbach:
> but actually you only need them to detect the trailing spaces. Otherwise you can use the tokenN position, the tokenN length and the tokenN+1 position to detect missing whites after if, foreach and such tokens.
> Also take care of the comments, since if/*bla bla*/(true) is perfectly valid.

Hmm I am still not sure how I could do this without access to the raw code. Allman is probably easy to look at as an example, in particular this method is the core of the Allman check:

```d
/**
Checks whether a brace or newline comes first
*/
void findBraceOrNewLine(size_t start, size_t end, size_t line, size_t column)
{
	import std.algorithm : canFind;
	import std.utf : byCodeUnit;

	auto codeRange = (cast(char[]) code[start..end]).byCodeUnit;

	// inline statements are allowed -> search for newline
	if (codeRange.canFind('\n'))
	{
		foreach (s; codeRange)
		{
			// first brace
			if (s == '{')
			{
				// DoStatement hasn't a proper line and column attached
				// -> calculate ourselves
				if (line == 0 && column == 0)
				{
					// find line & column of brace
					auto t = findLineAndColumnForPos(start);
					line = t.line + 1; // Dscanner starts lines at 1
					column = t.column;
				}
				addErrorMessage(line, column, KEY, MESSAGE);
				break;
			}
			// newline - test passed
			else if (s == '\n')
			{
				break;
			}
		}
	}
}
```

1) With the `code` array, I simply go back/forth because I know the `{start,end}Location` of an statement.  If I only have the `tokens` array, how could I access the next token without iterating through all of them?

2) And imho this check is rather easy to do with access to the code array and it's in memory anyways, so I don't really get why it's so bad to use it?

BBasile:
Allman enforcer:

```d
void main()
{
    import dparse.lexer;
    import std.stdio, std.file : read;

    ubyte[] src = cast(ubyte[]) read(__FILE_FULL_PATH__);
    StringCache cache = StringCache(StringCache.defaultBucketCount);
    LexerConfig config = LexerConfig("", StringBehavior.source);

    auto toks = getTokensForParser(src, config, &cache);

    foreach(const i; 1..toks.length) {
        if (toks[i].type == tok!"{" && toks[i-1].line == toks[i].line)
            writeln(__FILE_FULL_PATH__, "(", toks[i].line, "): Allman not respected !");
    }
}
```

could be nicer with the sliding window range (needs two token at the same time to write a pipeline)

wilzbach:
This is superseded by #446, #447 and #448.

PR #371, opened 2016-08-29
add detection of undocumented enums and their members

BBasile:
Maybe there's a specific reason why they're were not added in first place ? I think that even if the member names are often explicit it cannot tells everything.

### DIFF
--- a/src/analysis/undocumented.d
+++ b/src/analysis/undocumented.d
@@ -133,7 +133,9 @@ class UndocumentedDeclarationCheck : BaseAnalyzer
 	{
 	}
 
+	mixin V!AnonymousEnumMember;
 	mixin V!ClassDeclaration;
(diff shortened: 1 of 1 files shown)


src/analysis/undocumented.d
```diff
@@ -133,7 +133,10 @@ class UndocumentedDeclarationCheck : BaseAnalyzer
 	{
 	}
 
+	mixin V!AnonymousEnumMember;
 	mixin V!ClassDeclaration;
+	mixin V!EnumDeclaration;
+	mixin V!EnumMember;
```
Hackerpilot: I like the addition of AnonymousEnumMember and EnumDeclaration, but the check for enum members may not add as much value.

BBasile:
Ok, removed.

PR #372, opened 2016-09-02
add a check for the auto functions without return statement

BBasile:
Auto functions without return statement can be a sign of omission. The compiler deduces `void` as return type, which can lead to strange error messages when the return really misses.

example inspired by something that's really happened a few days ago:

```D
auto foo()
{
    asm{naked; mov RAX, 42; ret;}
}
```

This new _Dscanner_ check would have output:
> [warn]: Auto function without return statement, prefer an explicit void

### DIFF
--- a/src/analysis/auto_function.d
+++ b/src/analysis/auto_function.d
@@ -1,0 +1,87 @@
+//          Copyright Basile Burg 2016.
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file LICENSE_1_0.txt or copy at
+//          http://www.boost.org/LICENSE_1_0.txt)
+
+module analysis.auto_function;
+
+import analysis.base;
+import analysis.helpers;
+import dparse.ast;
+import dparse.lexer;
+
+import std.stdio;
+import std.algorithm.searching : any;
+
+/**
+ * Checks for auto functions without return statement.
+ *
+ * Auto function without return statement can be an omission and are not
+ * detected by the compiler. However sometimes they can be used as a trick
+ * to infer attributes.
+ */
+final class AutoFunctionChecker : BaseAnalyzer
+{
+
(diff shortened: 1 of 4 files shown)


src/analysis/config.d
```diff
@@ -140,8 +140,11 @@ struct StaticAnalysisConfig
 	string alias_syntax_check = Check.disabled;
 
 	@INI("Checks for else if that should be else static if")
-    string static_if_else_check = Check.disabled;
+	string static_if_else_check = Check.disabled;
```
UplinkCoder: @BBasile whitespace issue

BBasile: No, actually dscanner sources mostly use tab indentation. Here I've just replaced 4 spaces by a tab. In a diff GH displays 8 spaces for a tab instead of 4 (otherwise the diff could look like a "no diff") so it gives the feeling that there's an issue. You can [check the whole file](https://github.com/BBasile/Dscanner/blob/8f4ab6671a5bca8908d6a278bb77b89ea6075fb5/src/analysis/config.d).

src/analysis/auto_function.d
```diff
@@ -0,0 +1,87 @@
...
+		super(fileName, null, skipTests);
+	}
+
+	override void visit(const FunctionDeclaration decl)
+	{
+		_returns.length += 1;
+		scope(exit) _returns.length -= 1;
+		_returns[$-1] = false;
+
+		const bool autoFun = decl.storageClasses
```
BBasile: Q: Shouldn't `FunctionDeclaration.hasAuto` carry this information ? It seems to be never set.

Hackerpilot:
Looks good. Fix the merge conflicts and I'll merge this.

BBasile:
Conflict  resolved.

PR #374, opened 2016-09-08
Proposal: Add an alternative highlighter

Pursche:
The current highlighter outputs HTML styled code, I am working on D support for an IDE that doesn't use an HTML based text editor and would love to have an alternative highlighter that outputs something simpler. This is just a proposal to start a dialog and I am willing to refactor or rename anything in here.

The format I propose is as following:
*One token per line*
TYPE, Chars from start of file, Type/Text length

This format is extremely simple and powerful enough for us to highlight on, here's a printscreen of an example file and it's output: http://i.imgur.com/2GY74vl.png

In the future I would love if libdparse could detect that the second "other,26,3" was referring to the name "Foo" that I gave to my class.

### DIFF
--- a/src/althighlighter.d
+++ b/src/althighlighter.d
@@ -1,0 +1,88 @@
+//          Copyright Brian Schott (Hackerpilot) 2012.
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file LICENSE_1_0.txt or copy at
+//          http://www.boost.org/LICENSE_1_0.txt)
+
+module althighlighter;
+
+import std.stdio;
+import std.array;
+import std.conv;
+import std.string;
+import dparse.lexer;
+
+// http://ethanschoonover.com/solarized
+void highlight(R)(ref R tokens, string fileName)
+{
+	while (!tokens.empty)
+	{
+		auto t = tokens.front;
+		tokens.popFront();
+		if (isBasicType(t.type))
+		{
+			write("type");
+			writeln("," ~ to!string(t.index) ~ "," ~ to!string(str(t.type).length));
+		}
+		else if (isKeyword(t.type))
+		{
+			write("keyword");
+			writeln("," ~ to!string(t.index) ~ "," ~ to!string(str(t.type).length));
+		}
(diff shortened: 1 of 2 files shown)


src/althighlighter.d
```diff
@@ -0,0 +1,88 @@
...
+
+			text = strip(text);
+
+			if (text.length == 0)
+				continue;
+
+			write("other");
+			writeln("," ~ to!string(t.index) ~ "," ~ to!string(t.text.length));
+		}
+
```
BBasile: As a matter of style can you remove this huge gap ?

Hackerpilot:
If I understand this correctly, you want to use launch a bunch of D-Scanner processes as a user is typing in order to do syntax highlighting? That is never going to be fast.

Pursche:
@BBasile Absolutely.

@Hackerpilot Well the idea was to attempt to compile DScanner into a DLL, load it into the C# based editor as an unmanaged DLL and only run the highlighter on the actively edited parts. Do you have a better solution in mind?

Hackerpilot:
> Infact, I guess we could totally skip D-scanner and just link towards libdparser...

This is a much better idea.

BBasile:
>> Infact, I guess we could totally skip D-scanner and just link towards libdparser...

> This is a much better idea.

@Hackerpilot  The lexer you've made in libdparse must allow to parse so there's a token for everything: `tok!"++"`, `tok!"--"`, `tok!"abstract"`, etc. But the lexers used to highlight usually don't need as accurate informations. To highlight one just needs to know if it's an _operator_, a _symbol_, a _keyword_, etc but not which is the specific _operator_, _keyword_ or _symbol_.

So if you ( @Pursche ) need a D lexer in C# to highlight I'd suggest you to write one from scratch. It's not so hard. The only things that are a bit annoying with D are 
- all the string literal kinds
- Numbers literals are a bit too
- attributes: precise highlighting needs a bit of parsing because for example the keyword `const` can have 3 semantics ( :scream:) (depending on if it's a type constructor, parameter storage class or function attribute).

PR #379, opened 2016-11-26
auto functions checker, fix several cases of false warnings

BBasile:
prevent a few false warnings:
- `assert(0)`: function not designed to be called.
- `assert(false)`: function not designed to be called.
- `mixin("return" ...)`: try to find return in the literal.
- `extern(C) strlen(char*);` : so obvious, same for `@disable` 

more cases exist but they are not easily detectable.

Related to https://github.com/dlang/phobos/pull/4850#issuecomment-262364007.
@wilzbach, can you check if the cases that led you to disable the check are handled now ?

### DIFF
--- a/src/analysis/auto_function.d
+++ b/src/analysis/auto_function.d
@@ -29,6 +29,7 @@ private:
 	enum string MESSAGE = "Auto function without return statement, prefer an explicit void";
 
 	bool[] _returns;
+	size_t _mixinDepth;
 
 public:
 
@@ -40,7 +41,7 @@ public:
 		super(fileName, null, skipTests);
 	}
 
-	override void visit(const FunctionDeclaration decl)
+	override void visit(const(FunctionDeclaration) decl)
 	{
 		_returns.length += 1;
 		scope(exit) _returns.length -= 1;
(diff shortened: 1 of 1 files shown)


BBasile:
Note that it would be possible to skip `visit()` for a specific `version()` too. __Stddoc__ seemed to be a problem but the additions made for `assert()` should do the job.

wilzbach:
> @wilzbach, can you check if the cases that led you to disable the check are handled now ?

Wow that managed to catch a lot of false positives - great job :)
It seems we only have the mixin problem left:

```
std/range/package.d(8943:28)[warn]: Auto function without return statement, prefer an explicit void
std/range/package.d(8948:81)[warn]: Auto function without return statement, prefer an explicit void
std/range/package.d(8955:28)[warn]: Auto function without return statement, prefer an explicit void
std/range/package.d(8960:81)[warn]: Auto function without return statement, prefer an explicit void
std/range/package.d(9126:14)[warn]: Auto function without return statement, prefer an explicit void
std/range/package.d(9133:14)[warn]: Auto function without return statement, prefer an explicit void
std/experimental/ndslice/iteration.d(135:15)[warn]: Auto function without return statement, prefer an explicit void
std/experimental/ndslice/iteration.d(244:15)[warn]: Auto function without return statement, prefer an explicit void
```

There are just eight occurrences of mixin returns in Phobos, so it might be an option to use the slightly more verbose `return mixin(...)` alternative there.
Btw you can check the result yourself by running `make -f posix.mak style` (and setting the flag to enabled of course).

BBasile:
Thanks, yesterday I was browsing all the files and I didn't get where the hell dscanner was called.
I'll ping again If I manage to handle these cases.

wilzbach:
> I'll ping again If I manage to handle these cases.

I don't know whether this is possible because AFAIK is Dscanner unable to expand mixins and verify that they indeed have a `return` statement. It would be quite fancy to use the D Frontend or the upcoming CTFE engine, but I guess these are larger projects :/

wilzbach:
> I'll ping again If I manage to handle these cases.

Btw in case you need a reduced test case, here's a simple one:

```d
	assertAnalyzerWarnings(q{
		auto doStuff(){ mixin(_genSave());}
		private static string _genSave() @safe pure nothrow
        {
            return `import std.stdio;` ~
                   `return true`;
        }
	}, sac);
```

BBasile:
If the string is declared within the module that mixes it then it's possible to handle this case. But what will happen, like the previous changes, is not that the return is detected, it's more that a suspicious case of false detection is detected. This matches still to the policy of this check.

PR #381, opened 2016-12-07
fix #380, templated enum seen as unused variable

### LINKED ISSUE #380: templated enum seen as unused variable
```D
int main()
{
    enum a(T) = T.init;
    return a!int;
}
```

leads to a false detection.

BBasile:
test case:

```D
int main()
{
    enum a(T) = T.init;
    return a!int;
}
```


BBasile:
! Do not merge, better fix in progress !

BBasile:
ping, commit that fix 380 rebased with a better fix.

PR #383, opened 2016-12-10
prevent a case of false detection for auto functions

BBasile:
And that's the last one. For the others CTFE is needed. That's a bit the problem with template meta programming. So auto function check must be kept turned off for Phobos ;).

### DIFF
--- a/src/analysis/auto_function.d
+++ b/src/analysis/auto_function.d
@@ -30,6 +30,7 @@ private:
 
 	bool[] _returns;
 	size_t _mixinDepth;
+    string[] _literalWithReturn;
 
 public:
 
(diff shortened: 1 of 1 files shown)


src/analysis/auto_function.d
```diff
@@ -93,11 +94,63 @@ public:
...
+                    _returns[$-1] = true;
+        }
 	}
+
+    bool findReturnInLiteral(const(string) value)
+    {
+        import std.algorithm.searching : find;
+        import std.range : empty;
+
+        return value == "return" || !value.find("return ").empty;
```
BBasile: Despite of the complexity added for the last case, this will also be better on the previous code. I think it should handle full strings with a return , e.g `"expression; expression; return 0"` but also partially concatenated strings, e.g `return` ~ many other expressions....

PR #389, opened 2016-12-15
Add check for explictly annotated unittests

wilzbach:
Coming from https://github.com/dlang/phobos/pull/4952, this adds a style checker for explicitly annotated unittests.
For example these two example are properly annotated:

```d
@safe unittest {

}

@system unittest {

}
```

While for these two it will give a nice error message:

```d
unittest // [warn]: A unittest should be annotated with at least @safe or @system
{

}
pure nothrow @nogc unittest // [warn]: A unittest should be annotated with at least @safe or @system
{
}
```

### DIFF
--- a/src/analysis/explicitly_annotated_unittests.d
+++ b/src/analysis/explicitly_annotated_unittests.d
@@ -1,0 +1,116 @@
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file LICENSE_1_0.txt or copy at
+//          http://www.boost.org/LICENSE_1_0.txt)
+
+module analysis.explicitly_annotated_unittests;
+
+import dparse.lexer;
+import dparse.ast;
+import analysis.base : BaseAnalyzer;
+
+import std.stdio;
+
+/**
+ * Requires unittests to be explicitly annotated with either @safe or @system
+ */
+class ExplicitlyAnnotatedUnittestCheck : BaseAnalyzer
+{
+	enum string KEY = "dscanner.style.explicitly_annotated_unittest";
+	enum string MESSAGE = "A unittest should be annotated with at least @safe or @system";
+
+	///
+	this(string fileName, bool skipTests = false)
+	{
+		super(fileName, null, skipTests);
+	}
+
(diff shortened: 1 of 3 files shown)


src/analysis/explicitly_annotated_unittests.d
```diff
@@ -0,0 +1,116 @@
...
+	this(string fileName, bool skipTests = false)
+	{
+		super(fileName, null, skipTests);
+	}
+
+	override void visit(const Declaration decl)
+	{
+		if (decl.unittest_ !is null)
+		{
+			bool isSafeOrSystem = false;
```
BBasile: the initializer is not required here.

src/analysis/explicitly_annotated_unittests.d
```diff
@@ -0,0 +1,116 @@
...
+	import std.stdio : stderr;
+	import std.format : format;
+	import analysis.config : StaticAnalysisConfig, Check;
+	import analysis.helpers : assertAnalyzerWarnings;
+
+	StaticAnalysisConfig sac;
+	sac.explicitly_annotated_unittests = Check.enabled;
+
+	assertAnalyzerWarnings(q{
+		@safe unittest {
```
BBasile: can you change all the test to one line  `... unittest {}` ? (and also in the nested tests)

src/analysis/explicitly_annotated_unittests.d
```diff
@@ -0,0 +1,116 @@
...
+	{
+		if (decl.unittest_ !is null)
+		{
+			bool isSafeOrSystem = false;
+			if (decl.attributes !is null)
+			foreach (attribute; decl.attributes)
+			{
+				if (attribute.atAttribute !is null)
+				{
+					auto token = attribute.atAttribute.identifier.text;
```
BBasile: `const token = ...`,  you would get a  dscanner warning here.

wilzbach:
> Obviously useful for Phobos. Only a few minor details to change but otherwise it looks ok.

Thanks a lot for your quick review :)

andralex:
thx!

PR #390, opened 2016-12-19
Add Ddoc check for properly documented public functions

wilzbach:
From https://wiki.dlang.org/Vision/2016H2:

> Make sure every function in Phobos has Params: and Returns: sections http://www.digitalmars.com/d/archives/digitalmars/D/Phobos_Documentation_-_call_to_action_258777.html

For Phobos this returns atm 1.1K symbols, which then can be fixed and properly annotated.

This check could also be extended to check whether all parameters are correctly defined in the `Params` section.

Note that for simplicity this checks only targets public functions and ignores e.g. struct/class constructors (for now).

### DIFF
--- a/src/analysis/properly_documented_public_functions.d
+++ b/src/analysis/properly_documented_public_functions.d
@@ -1,0 +1,669 @@
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file LICENSE_1_0.txt or copy at
+//          http://www.boost.org/LICENSE_1_0.txt)
+
+module analysis.properly_documented_public_functions;
+
+import dparse.lexer;
+import dparse.ast;
+import analysis.base : BaseAnalyzer;
+
+import std.format : format;
+import std.range.primitives;
+import std.stdio;
+
+/**
+ * Requires each public function to contain the following ddoc sections
+	- PARAMS:
+		- if the function has at least one parameter
+		- every parameter must have a ddoc params entry (applies for template paramters too)
+		- Ddoc params entries without a parameter trigger warnings as well
+	- RETURNS: (except if it's void, only functions)
+ */
+class ProperlyDocumentedPublicFunctions : BaseAnalyzer
+{
+	enum string MISSING_PARAMS_KEY = "dscanner.style.doc_missing_params";
+	enum string MISSING_PARAMS_MESSAGE = "Parameter %s isn't documented in the `Params` section";
+	enum string MISSING_TEMPLATE_PARAMS_MESSAGE
+		= "Template parameters %s isn't documented in the `Params` section";
+
+	enum string NON_EXISTENT_PARAMS_KEY = "dscanner.style.doc_non_existing_params";
+	enum string NON_EXISTENT_PARAMS_MESSAGE = "Ddoced parameter %s isn't a function parameter";
+
+	enum string MISSING_RETURNS_KEY = "dscanner.style.doc_missing_returns";
+	enum string MISSING_RETURNS_MESSAGE = "A public function needs to contain a `Returns` section";
+
+	///
+	this(string fileName, bool skipTests = false)
+	{
+		super(fileName, null, skipTests);
+	}
+
+	override void visit(const Module mod)
+	{
+		islastSeenVisibilityLabelPublic = true;
+		mod.accept(this);
+		postCheckSeenDdocParams();
+	}
+
+	override void visit(const Declaration decl)
+	{
+		import std.algorithm.searching : any;
+		import std.algorithm.iteration : map;
+
+		// skip private symbols
+		enum tokPrivate = tok!"private",
+			 tokProtected = tok!"protected",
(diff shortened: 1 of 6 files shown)


src/analysis/properly_documented_public_functions.d
```diff
@@ -0,0 +1,202 @@
...
+		mod.accept(this);
+	}
+
+	override void visit(const Declaration decl)
+	{
+		import std.algorithm.searching : any;
+		import std.algorithm.iteration : map;
+
+		// skip private symbols
+		enum tokPrivate = tok!"private", tokProtected = tok!"protected", tokPackage = tok!"package";
```
BBasile: Q: Did you choose explicitly to analyze `export` functions ?

src/analysis/properly_documented_public_functions.d
```diff
@@ -0,0 +1,202 @@
...
+	enum string MISSING_RETURNS_KEY = "dscanner.style.doc_missing_returns";
+	enum string MISSING_RETURNS_MESSAGE = "A public function needs to contain a `Returns` section";
+
+	///
+	this(string fileName, bool skipTests = false)
+	{
+		super(fileName, null, skipTests);
+	}
+
+	override void visit(const Module mod)
```
wilzbach: Are you sure?

It is implemented [here](https://github.com/Hackerpilot/libdparse/blob/master/src/dparse/ast.d#L260) and the compiler tells me so:

```
src/analysis/properly_documented_public_functions.d(32): Error: cannot implicitly override base class method dparse.ast.ASTVisitor.visit with analysis.properly_documented_public_functions.ProperlyDocumentedPublicFunctions.visit; add 'override' attribute
```

src/analysis/properly_documented_public_functions.d
```diff
@@ -0,0 +1,202 @@
...
+
+		if (decl.functionDeclaration !is null)
+			decl.accept(this);
+	}
+
+	override void visit(const FunctionDeclaration decl)
+	{
+		import std.algorithm.searching : canFind;
+
+		// ignore header declaration for now
```
wilzbach: I am not sure on this. The motivation for now was just to get less errors with `Druntime`, but thinking more about it, I think that also these definition should be properly documented.

src/analysis/properly_documented_public_functions.d
BBasile: Actually I can guess why you've put this useless overridden function: you wished to check that each module declaration has a documentation comment and finally you've decided to restrict the feature to functions.

src/analysis/properly_documented_public_functions.d
wilzbach: No, this was just a "hack" to only look at publicly exposed functions. I added another check to look out for `private` etc. labels.

Hackerpilot:
Have you considered using https://github.com/economicmodeling/libddoc/ to check the contents of the doc comments? Its code is going to be more exact than `canFind` and it'll help you more easily determine if there are undocumented arguments (or documented arguments that don't exist).

wilzbach:
> Have you considered using https://github.com/economicmodeling/libddoc/ to check the contents of the doc comments?

I simply didn't know about `libddoc` ;-)

I rebased this PR to add `libddoc` as git and dub dependency.
Moreover I added:
- a check to ensure that every function/template parameter has an respective ddoc entry and reversely that there's no ddoc parameter that doesn't match a function/template parameter.
- checks for structs, classes, templates, ...
- removed the skip checking over header-only declarations
- more tests

src/analysis/properly_documented_public_functions.d
```diff
@@ -0,0 +1,669 @@
...
+		- every parameter must have a ddoc params entry (applies for template paramters too)
+		- Ddoc params entries without a parameter trigger warnings as well
+	- RETURNS: (except if it's void, only functions)
+ */
+class ProperlyDocumentedPublicFunctions : BaseAnalyzer
+{
+	enum string MISSING_PARAMS_KEY = "dscanner.style.doc_missing_params";
+	enum string MISSING_PARAMS_MESSAGE = "Parameter %s isn't documented in the `Params` section";
+	enum string MISSING_TEMPLATE_PARAMS_MESSAGE
+		= "Template parameters %s isn't documented in the `Params` section";
```
Hackerpilot: Please add periods to the ends of the messages.

BBasile:
About the missing return section, How `auto` functions are handled ?

wilzbach:
> It is confused by the following (slightly unusual) code though:

Do I understand it correctly that all parameters should be "grouped" here? If that's so, it might not be that hard to support as a simple flag to keep track of this nested "ditto" state could be enough.

> About the missing return section, How auto functions are handled ?

Atm like any other return type. I added a small test case for it.
There's any "smart" feature implemented that parses the declarations and checks the type of the `return` statements to detect a `auto` function that is `void`. However I think there's another Dscanner check that does exactly this (warns when `auto` is used instead of `void`)

Issue #128, opened 2014-02-25
Warn about braces not on their own line


Hackerpilot:
Phobo style guide requires braces to be on their own line.

wilzbach:
Fixed with https://github.com/dlang-community/D-Scanner/pull/446

PR #394, opened 2016-12-26
If constraints: check for same indent level

wilzbach:
See also: https://github.com/dlang/dlang.org/pull/1534

This is based on https://github.com/Hackerpilot/Dscanner/pull/348 as the `sourceCode` is needed to count the whitespace tokens at the  beginning of the line.

### DIFF
--- a/src/analysis/allman.d
+++ b/src/analysis/allman.d
@@ -1,0 +1,221 @@
+// Distributed under the Boost Software License, Version 1.0.
+//	  (See accompanying file LICENSE_1_0.txt or copy at
+//			http://www.boost.org/LICENSE_1_0.txt)
+
+module analysis.allman;
+
+import dparse.lexer;
+import dparse.ast;
+import analysis.base : BaseAnalyzer;
+import dsymbol.scope_ : Scope;
+
+/**
+Checks for the allman style (braces should be on their own line)
+
+------------
+if (param < 0) {
+
+}
+------------
+
+should be
+
+------------
+if (param < 0)
+{
+
+}
+------------
+*/
+class AllManCheck : BaseAnalyzer
+{
+	///
+	this(string fileName, const(ubyte)[] code, bool skipTests = false)
+	{
+		super(fileName, null, skipTests);
+		this.code = code;
+	}
+
+	override void visit(const WhileStatement st)
+	{
+		if (st.declarationOrStatement !is null)
+			checkForBrace(st.declarationOrStatement, st.expression.line, st.expression.column);
+	}
+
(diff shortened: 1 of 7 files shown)


andralex:
I see you also made some refactorings here - nice.

andralex:
cch @Hackerpilot :)

wilzbach:
> I see you also made some refactorings here - nice.

This is only #348 as this needs access to the raw `sourceCode` array to count the indentation level.

wilzbach:
@Hackerpilot I am a bit stuck here.
As you stated on #348 you would prefer not to expose the raw source code to the plugins, which is 
understandable. However, with only the token array at hand, how can I look ahead/back to the next/previous token (apart from iterating through the tokens array all the time)?
From what I [learned](https://github.com/Hackerpilot/libdparse/blob/master/src/std/experimental/lexer.d#L284) the token doesn't know its position within the array, it only knows about:

```
type = the token type
text = the text of the token, which may be null
line = the line number at which this token occurs
column = the column number at which this token occurs
index = the byte offset from the beginning of the input at which this token occurs
```

Or to be more precise, for the indentation check the plugin looks at a couple of *Declarations in the AST. Then it compares the indent level of its constraint with the indent of the name token itself. For example for `FunctionDeclaration`:

```d
override void visit(const FunctionDeclaration decl) {
	if (decl.constraint !is null)
		checkConstraintSpace(decl.constraint, decl.name);
}
```

Currently `checkConstraintSpace` is done by looking back until the beginning of the line and counting the seen whitespace characters, which is the indented level of the token:

```d
auto lineIndent = code[0 .. constraint.location].retro
				.until('\n')
				.walkLength;

auto tokenLineStart = code[0 .. location].retro
						.until('\n')
						.walkLength;

auto tokenIndent = code[location - tokenLineStart .. location]
					.until!(x => !x.isWhite)
					.walkLength;

if (tokenIndent != lineIndent)
	addErrorMessage(constraint.expression.line, constraint.expression.column, KEY, MESSAGE);
```

How would I do this with merely the token array? Or can I (as proposed before) just pass the `code` array into the plugin (it's already in memory anyways).

BBasile:
@wilzbach
- you should also add a line to the _readme.md_, so far each check is briefly described.
- this PR actually fixes #128, so you should add `fix #128 - ` as prefix in the commit message.

andralex:
@wilzbach @Hackerpilot I see this needs some rebasing, right?

wilzbach:
> @wilzbach @Hackerpilot I see this needs some rebasing, right?

Superseded by #450

PR #412, opened 2017-02-23
disable Phobos-specific checks by default, close #406

### LINKED ISSUE #406: add a --skipPhobosChecks switch
With the recent additions related to phobos a lot of noise is generated for the import sortedness and the unittest annotations. To disable them a config file is required. To continue using dscanner without config file this switch seems necessary. (this was also the idea behind the --slipTests switch).

So this switch would disable the checks made especially for phobos CI.

BBasile:
The recent checks added by @wilzbach were specifically made for Phobos and they tend to generate a lot of noise in user code, where the style rules may be less strict.

the PR
- set accordingly the default values in the struct.
- removes `fillConfig` and instead the default values of the struct are used.
- a special function is used to apply the `--skipTest` command line option.
- finally unittest are modified to handle the new initilization scheme (will break opened PRs that add new checks).

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,9 @@ Note that the "--skipTests" option is the equivalent of changing each
 * Some assertions that check conditions that will always be true.
 * Auto functions without return statement. The compiler doesn't see an omission and it infers 'void' as return type.
 * `final` attribute is used but in this context it's a noop.
+* Check for properly documented public functions ("Returns" and "Params" sections). Initially implemented to lint Phobos. By default disabled.
+* Check for explicitly annotated unittests (_@system_ or _@safe_). Initially implemented to lint Phobos. By default disabled.
+* Check for that imports are sorted. Initially implemented to lint Phobos. By default disabled.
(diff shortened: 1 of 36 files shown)


BBasile:
Bad, luck Travis for D seems to be broken. However the tests pass locally.

rikkimax:
@Hackerpilot can you please take a look at this PR? I will merge it in a couple of days unless you comment on it.

BBasile:
Thanks. 

In case anyone worked (i think especially to @wilzbach pending PRs) on a new check
- in the unittests, initialize `sac` with `disabledConfig()`
- in `analysis.config`, set the new check to `Check.enabled` unless this is a very phobos specific check, in which case you'll have to set it to `Check.disabled`.

wilzbach:
> In case anyone worked (i think especially to @wilzbach pending PRs) on a new check

Thanks a lot for the info - will have a look at rebasing the pending PRs when time allows ;-)

PR #419, opened 2017-03-03
Update build.bat

Zoadian:
added missing libddoc to build.bat


wilzbach:
This is now the third PR for this in the queue :O

PR #427, opened 2017-05-07
update readme badge address

BBasile:


wilzbach:
As this is the first precedent for a trivial change I would like to wait for @Hackerpilot's opinion on policies or guidelines that we should follow? For example, should such simple changes be squashed into one commit (this can be done via GH)?

BBasile:
I edited online, in the opposite case i would have marked the second as fixup. There's no problem here, just squash, it's the obvious thing to do.

wilzbach:
> I edited online, in the opposite case i would have marked the second as fixup. There's no problem here, just squash, it's the obvious thing to do.

Yep you are right and I will do so once @Hackerpilot shared his thoughts on the future of his repos.
As mentioned this is the first precedent and imho it's better to be safe than sorry even though I agree that this is absolutely trivial.

Hackerpilot:
> For example, should such simple changes be squashed into one commit (this can be done via GH)?

I don't have a string opinion either way.

PR #428, opened 2017-05-08
fix #352 - False positive: Parameter is never used, pointers

### LINKED ISSUE #352: False positive: Parameter is never used
void f(int *x) { *x = 1; }

BBasile:


BBasile:
There 's a problem, this PR is still visible on the dashboard https://github.com/pulls?user=dlang-community. If anyone knows why please explain !

PR #429, opened 2017-05-09
Add a check for useless initializers, close #285

### LINKED ISSUE #285: Warn when unnecessary initial values are used
e.g.

```
int i = 0;
double j = double.nan;
S some_struct = S();
```

should produce a warning.

BBasile:


wilzbach:
Hmm there are 513 occurrences for this on Phobos: http://sprunge.us/cAVH 

So I guess if there's no automated fixer [like dscanner-fix](https://github.com/wilzbach/dscanner-fix), we probably can't enable this check :(

BBasile:
he he. 513, that's funny how people tends to write initializers even if  they know they are useless, "just in case of".

PR #430, opened 2017-05-11
Refactor travis config to include osx, dmd-beta, and dub.

skl131313:
There a problem building inifiled in release mode, so will need to wait on that to get fixed before this is green.

https://github.com/burner/inifiled/issues/8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 sudo: false
 language: d
-script:
-    - git submodule update --init --recursive
-    - make test
+d:
+  - dmd-beta
+  - dmd
+  - ldc
+os:
+  - linux
+  - osx
+
+script: ./.travis.sh
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+dub build --build=release
+git submodule update --init --recursive
+
+if [ $DC = ldc2 ]; then
+    make test DC=ldmd2
+else
+    make test
+fi


.travis.yml
```diff
@@ -1,5 +1,14 @@
 sudo: false
 language: d
+d:
+  - dmd-beta
+  - dmd
```
wilzbach: how about adding LDC?

skl131313: Would need to change the makefile and how testing is done to support it. Possibly by adding a testing into the dub config.

wilzbach: Hmm, DUB will automatically detect that only ldc is available and use it to run the tests ;-)
In the Makefile `DC ?= dmd` will mean that it can be overwritten by the environment variable, so it should work out of the box.
However, feel free to add this in a further step.

BBasile:
Finally, what's up with inifiled ? Shouldn't the version be changed in the DUB description ?

skl131313:
@BBasile I could do a separate PR for that, this PR strictly changes the travis build to include dub to ensure it can build through it, as well as adding more configurations.

There was a bug in inifiled v1.0.0 (the latest at the time of creating this PR, there's 1.0.1 now) that caused a build error in release mode. As far as I know the interface hasn't changed and it could very well still compile with 0.6.0, I'd need to try it.

BBasile:
Right, no need to upgrade. Sorry it was a misunderstanding of what dub does with `=>`.

PR #432, opened 2017-05-12
fix #431, Virtual call in ctor - problem with overloaded virtual methods

### LINKED ISSUE #431: Virtual call in ctor - problem with overloaded virtual methods
A warning is emitted for each overload of the virtual method called in the constructor.
This is way too verbose, e.g with the visitor pattern.

BBasile:
Also enable the check, which was not by default.


BBasile:
Indeed this is trivial, just a detail i omitted in the initial PR.

PR #441, opened 2017-06-12
Add check for redundant attributes (for now: redundant access specifier)

wilzbach:
When I started to write this check, it was mainly to avoid redundant access specifiers like:

```d
private:
   private int foo;
```

However, I realized that it can work with any kind of attribute, e.g.

```d
@safe:
   @safe void foo() {}
```

or 

```
@myCustomUDA {
    @myCustomUDA void foo() {}
} 
```

However, for now I lack the time to properly test all attribute combination, so I wrote generic code, but only enabled it for access specifier. It shouldn't be too hard to extend it to all attributes later.

### DIFF
--- a/src/analysis/redundant_attributes.d
+++ b/src/analysis/redundant_attributes.d
@@ -1,0 +1,290 @@
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file LICENSE_1_0.txt or copy at
+//          http://www.boost.org/LICENSE_1_0.txt)
+
+module analysis.redundant_attributes;
+
+import dparse.ast;
+import dparse.lexer;
+import dsymbol.scope_ : Scope;
+import analysis.base;
+import analysis.helpers;
+
+import std.algorithm;
+import std.conv : to, text;
+import std.range : empty, front, walkLength;
+
+/**
+ * Checks for redundant attributes. At the moment only visibility attributes.
+ */
+class RedundantAttributesCheck : BaseAnalyzer
+{
+	this(string fileName, const(Scope)* sc, bool skipTests = false)
+	{
+		super(fileName, sc, skipTests);
+		stack.length = 0;
+	}
+
+	override void visit(const Declaration decl)
+	{
+
+		// labels, e.g. private:
+		if (auto attr = decl.attributeDeclaration)
+		{
+			if (filterAttributes(attr.attribute))
+			{
+				addAttribute(attr.attribute);
+			}
+		}
+
+		auto attributes = decl.attributes.filter!(a => filterAttributes(a));
+		if (attributes.walkLength > 0) {
+
+			// new scope: private { }
+			if (decl.declarations.length > 0)
+			{
+				auto prev = currentAttributes[];
+				// append to current scope and reset once block is left
+				foreach (attr; attributes)
+					addAttribute(attr);
+
+				scope(exit) currentAttributes = prev;
+				decl.accept(this);
+			} // declarations, e.g. private int ...
+			else
+			{
+				foreach (attr; attributes)
+					checkAttribute(attr);
+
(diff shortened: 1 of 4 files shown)


BBasile:
I'd like to see a test for

```d
private:
public int a;
private: // should warn here
```

src/analysis/redundant_attributes.d
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	bool checkAttribute(const Attribute attr)
+	{
+		auto match = currentAttributes.find!(a => a.attribute.type == attr.attribute.type);
+		if (!match.empty)
+		{
+			auto token = attr.attribute;
+			addErrorMessage(token.line, token.column, KEY,
+					text("same access specifier used as defined on line ",
```
BBasile: "access specifier" is inaccurate, it's called a [visibility attribute](https://dlang.org/spec/attribute.html#VisibilityAttribute).
If you add support for other attribute you'll have to think to use the correct term too.

src/analysis/redundant_attributes.d
```diff
@@ -0,0 +1,267 @@
...
+	alias CurrentScope = ConstAttribute[];
+	ref CurrentScope currentAttributes() {
+		return stack[$ - 1];
+	}
+	CurrentScope[] stack;
+
+	void addAttribute(const Attribute attr)
+	{
+		removeOverwrite(attr);
+		if (checkAttribute(attr)) {
```
BBasile: please don't blend the brace styles. same remark plenty of times !

BBasile:
small remark for the future: you can push the feature branch directly on dlang-community. It's easier for the reviewers to pull the branch on their clone.

wilzbach:
> I'd like to see a test for

Added. Btw https://github.com/dlang/phobos/pull/5477 contains all found instances in Phobos, but as mentioned it doesn't check all attributes (for now).

> style + use same name as in the grammar

Fixed. Sorry.

> small remark for the future: you can push the feature branch directly on dlang-community. It's easier for the reviewers to pull the branch on their clone.

Okay. I will do so next time.
Btw you can rather easily pull a PR with this git config:

```
[remote "upstream"]
	fetch = +refs/pull/*/head:refs/remotes/upstream/pr/*
```
(or `master` if you don't use an upstream here).

and then e.g. `git checkout pr/441`

BBasile:
I suspected the current failure. It's on the test i asked you to add.

wilzbach:
> I suspected the current failure. It's on the test i asked you to add.

Hmm are you sure?
AFAICT I just forgot to adjust the line numbers after changing the brace style?

BBasile:
> AFAICT I just forgot to adjust the line numbers after changing the brace style?

Right.

Appveyor icon is still there ?

wilzbach:
> Appveyor icon is still there ?

It will just be gone for now PRs ...

wilzbach:
(rebased to master)

BBasile:
Do you want to use this on phobos ASAP ? For me it's quite mergeable in the current state. Also you didn't touch it last 2 days, which means that it's also okay for you.

wilzbach:
> Do you want to use this on phobos ASAP ?

I currently have other goals, but eventually at some point in the future, yes.
But I haven't had time to fix https://github.com/dlang/phobos/pull/5477, so there's no need to hurry.
I just saw the merge error here ...

> For me it's quite mergeable in the current state. Also you didn't touch it last 2 days, which means that it's also okay for you.

I eventually want to make it work with all kind of attributes, but unfortunately not in this month or in the foreseeable future.

BBasile:
> I eventually want to make it work with all kind of attributes

There's chance that all attribute require much more work and will not be compatible with the current structure of this check. IIRC for a `@safe struct{}` safety is propagated to functions templates too.

PR #446, opened 2017-06-12
Add allman check

wilzbach:
Thanks a lot to @BBasile for this awesomely [simple idea](https://github.com/dlang-community/D-Scanner/pull/348#issuecomment-306937458).


wilzbach:
Btw this doesn't find any violation at Phobos and in theory it can remove [two of the ugly RegExp](https://github.com/dlang/phobos/blob/master/posix.mak#L550).

PR #449, opened 2017-06-12
Allow duplicate braces in allman style

wilzbach:
It seems like I have tested Phobos yesterday with a wrong dscanner binary.
However, there's only one small need for amendment that I found.

Apparently e.g in `std/experimental/allocator/common.d` this brace style seems to be popular for `static if`:

```d
        static if (hasMember!(A, "alignedAllocate"))
        {{
             auto b3 = a.alignedAllocate(1, 256);
             assert(b3.length <= 1);
             assert(b3.ptr.alignedAt(256));
             assert(a.alignedReallocate(b3, 2, 512));
             assert(b3.ptr.alignedAt(512));
             static if (hasMember!(A, "alignedDeallocate"))
             {
                 a.alignedDeallocate(b3);
             }
        }}
```

### DIFF
--- a/src/analysis/allman.d
+++ b/src/analysis/allman.d
@@ -40,12 +40,18 @@ class AllManCheck : BaseAnalyzer
 				// ignore struct initialization
 				if (tokens[i-1].type == tok!"=")
 					continue;
+				// ignore duplicate braces
+				if (tokens[i-1].type == tok!"{" && tokens[i - 2].line != curLine)
+					continue;
 				// ignore inline { } braces
 				if (curLine != tokens[i + 1].line)
 					addErrorMessage(tokens[i].line, tokens[i].column, KEY, MESSAGE);
 			}
 			if (tokens[i].type == tok!"}" && curLine == prevTokenLine)
 			{
+				// ignore duplicate braces
+				if (tokens[i-1].type == tok!"}" && tokens[i - 2].line != curLine)
+					continue;
 				// ignore inline { } braces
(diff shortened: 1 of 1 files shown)


wilzbach:
FYI in case you are interested: the PR at Phobos to prepare for enabling the allman check: https://github.com/dlang/phobos/pull/5480

BBasile:
what's the point with this thing ? Don't know how to call it btw _immediate scope_ ?

wilzbach:
> what's the point with this thing ?

It's a "pattern" that appears in Phobos ...

> Don't know how to call it btw immediate scope ?

Good idea.

Btw we seem to have an issue with the non-DUB builds & dmd-nightly:

```
containers/src/containers/treemap.d(33): Error: field tree must be initialized in constructor
containers/src/containers/treemap.d(166): Error: template instance containers.treemap.TreeMap!(int, int, StatsCollector!(FreeList!(AllocatorList!(Factory, GCAllocator), 64LU, 64LU, cast(Flag)false), 262143LU, 0LU)*, "a < b", false, 64LU) error instantiating
```

I think DUB managed to evade this by not trying to build it and just passing the containers as input directory?

wilzbach:
> Btw we seem to have an issue with the non-DUB builds & dmd-nightly:

See also: https://github.com/economicmodeling/containers/issues/80

(it's a regression from https://github.com/dlang/dmd/pull/6816, similarly to https://github.com/dlang-community/D-YAML/issues/66)

BBasile:
based on the previous comment i guess this is ok to merge.

PR #451, opened 2017-06-13
update libdparse

BBasile:
There's a deprecation message when it's build with dmd-nightly. It matches to something that's fixed in the latest dparse version. The version used by D-Scanner was ~~quite~~ a bit outdated anyway.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -10,7 +10,7 @@
 		"StdLoggerDisableWarning"
 	],
 	"dependencies": {
-		"libdparse": "~>0.7.0",
+		"libdparse": "~>0.7.1-beta.2",
 		"dsymbol": "~>0.2.0",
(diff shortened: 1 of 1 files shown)


wilzbach:
Travis failure is unrelated, so I am using the administrative power to merge this until the regression is fixed or worked around.

dub.json
```diff
@@ -10,7 +10,7 @@
 		"StdLoggerDisableWarning"
 	],
 	"dependencies": {
-		"libdparse": "~>0.7.0",
+		"libdparse": "~>0.7.1-beta.2",
```
wilzbach: Btw normally the `dub.selections.json` is commited as well. Then there's no need to modify the `dub,json` file ...

PR #458, opened 2017-06-14
Fix deprecations

wilzbach:
They were getting annoying ;-)

### DIFF
--- a/src/analysis/useless_assert.d
+++ b/src/analysis/useless_assert.d
@@ -14,10 +14,10 @@ import std.stdio;
 
 auto filterChars(string chars, S)(S str)
 {
-	import std.algorithm.comparison : among;
+    import std.algorithm.comparison : among;
     import std.algorithm.iteration : filter;
     import std.meta : aliasSeqOf;
-	return str.filter!(c => !c.among(aliasSeqOf!chars));
+    return str.filter!(c => !c.among(aliasSeqOf!chars));
 }
 
 /**
--- a/src/analysis/useless_assert.d
+++ b/src/analysis/useless_assert.d
@@ -12,6 +12,14 @@ import dparse.lexer;
 
 import std.stdio;
 
+auto filterChars(string chars, S)(S str)
+{
+	import std.algorithm.comparison : among;
+    import std.algorithm.iteration : filter;
+    import std.meta : aliasSeqOf;
+	return str.filter!(c => !c.among(aliasSeqOf!chars));
+}
+
 /**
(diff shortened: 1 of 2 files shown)


src/analysis/useless_assert.d
```diff
@@ -42,11 +43,11 @@ class UselessAssertCheck : BaseAnalyzer
 		if (!skipSwitch) switch (token.type)
 		{
 		case tok!"doubleLiteral":
-			if (!token.text.removechars("Ll").to!double)
+			if (!token.text.asLowerCase.filter!(c => c != 'l').to!double)
```
ZombineDev: I think the following may be shorter/faster:
```D
auto filterChars(string chars, S)(S str)
{
	import std.algorithm.comparison : among;
    import std.algorithm.iteration : filter;
    import std.meta : aliasSeqOf;
	return str.filter!(c => !c.among(aliasSeqOf!chars));
}

if (!token.text.filterChars!"Ll".to!double)
    // ...

```

wilzbach:
> I think the following may be shorter/faster:

Yes, that's very elegant!

ZombineDev:
(Don't mind my stupid ws commits, I'll squash them out of existance as soon as the tests pass ;) )

wilzbach:
> (Don't mind my stupid ws commits, I'll squash them out of existance as soon as the tests pass ;) )

No worries, I also made the same mistake here. Setting up a `.editorconfig` plugin in your favorite IDE can really help ;-)
Btw don't mind the test failure on dmd-nightly for now - it's a [known](https://issues.dlang.org/show_bug.cgi?id=17506) regression in EMSI containers.

ZombineDev:
>No worries, I also made the same mistake here. Setting up a .editorconfig plugin in your favorite IDE can really help ;-)

Ah yes, I have done so ages ago, but sometimes I'm too lazy and I edit PR with the GitHub online editor.

> Btw don't mind the test failure on dmd-nightly for now - it's a known regression in EMSI containers.

I noticed, but I can't make sense of the error - I don't think it should be there. Anyway, I think we should also bring emsi_containers to the org, so we can more easily manage such issues.

wilzbach:
> I noticed, but I can't make sense of the error - I don't think it should be there.

It was introduced here: https://github.com/dlang/dmd/pull/6816

>  Anyway, I think we should also bring emsi_containers to the org, so we can more easily manage such issues.

That's a great idea -> https://github.com/economicmodeling/containers/issues/81

PR #459, opened 2017-06-15
Workaround 17506 by excluding treemap.d

wilzbach:
In theory this should fix Travis for now. An alternative would be to use `allow_failures`


wilzbach:
> If you add a comment and open an issue as reminder to remove the workaround when dmd-nightly is fixed, I think we can merge this.

Done -> https://github.com/dlang-community/D-Scanner/issues/461

PR #462, opened 2017-06-16
set more def bucket count to get better perfs on huge modules

BBasile:
This is related to https://github.com/dlang-community/libdparse/issues/147.
Improvement is verified on prime_4m.d.


BBasile:
To verify `time dscanner primes_4m.d --sloc`, before and after. I cant obviously test this in dscanner itself because of the module size.

BBasile:
trivial

PR #463, opened 2017-06-18
set more def bucket count to get better perfs on huge modules

BBasile:


ZombineDev:
Perhaps we should change `StringCache.defaultBucketCount`?

ZombineDev:
It would be nice to also add a shorter version of the primes test case and assert that it takes less than a second.

BBasile:
> Perhaps we should change StringCache.defaultBucketCount ?

I have the whole code.dlang.org data locally. I thing i'll compute the average number of tokens per module (only those with string cached, like `tok!"identifier"`)  (note: there are around ~22000 modules). Then the right number will be its `nextPow2()`.

src/main.d
```diff
@@ -169,7 +169,7 @@ else
 
 	immutable usingStdin = args.length == 1;
 
-	StringCache cache = StringCache(StringCache.defaultBucketCount);
+	StringCache cache = StringCache(0x10000);
```
wilzbach: Whatever this number will be, imho it should definitely be accompanied by an explanatory comment ;-)

BBasile:
With the time passing i'm less concerned by this issue. There's still a trace of the problem in libdparse issues.

PR #464, opened 2017-06-18
Revert "Workaround 17506 by excluding treemap.d (#459)"

ZombineDev:
This reverts commit 80b831addaece9f0c4f65b27eb501767db1f7fb1.
The workaround is no longer necessary as the problematic
feature has been reverted from dmd-nightly.


wilzbach:
Thanks!

PR #465, opened 2017-06-19
Fix line_length checker for multiLine literals

wilzbach:
This is one of the few checks where access to the raw source code would make it a lot easier.

Note that this still doesn't fix the checker entirely as lines containing tabs will have a lower line length as libdparse counts columns bytewise. However, there's no easy way around this with no access to the raw source (according to previous discussions this isn't wanted)

In any case this fixes the false positives in Phobos.
The false negatives (with tabs) are a lot less harmless and were totally ignored before.

### DIFF
--- a/src/analysis/line_length.d
+++ b/src/analysis/line_length.d
@@ -26,25 +26,24 @@ class LineLengthCheck : BaseAnalyzer
 
 	override void visit(const Module)
 	{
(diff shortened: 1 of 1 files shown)


BBasile:
If only people could understand that their tab, from the editor point of view, is only a problem...

PR #466, opened 2017-06-19
Fix AssertError on unknown label in unused_label

wilzbach:
I tried to run Dscanner with all checks enabled (aka empty config) on Phobos (this might be a good CI check).

For now, I simply inserted a TODO when a unknown label is used, but I will open a new issue, s.t. it's not forgotten.

### DIFF
--- a/src/analysis/unused_label.d
+++ b/src/analysis/unused_label.d
@@ -143,8 +143,11 @@ private:
 	{
 		foreach (label; current.byValue())
 		{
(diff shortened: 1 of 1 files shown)


wilzbach:
> For now, I simply inserted a TODO when a unknown label is used, but I will open a new issue, s.t. it's not forgotten.

https://github.com/dlang-community/D-Scanner/issues/467

PR #470, opened 2017-06-20
Add has_public_example check

wilzbach:
This adds a check for public declaration without a documented unittest.

As Dscanner will hopefully support selective imports [soon](https://github.com/dlang-community/D-Scanner/pull/460), this is intended to be a replacement for the version currently used by Phobos (e.g. https://github.com/dlang/tools/blob/master/styles/has_public_example.d, https://github.com/dlang/phobos/blob/548630d2a9ba95f079193d7b47ed8cd685c0f7c8/posix.mak#L602).

### DIFF
--- a/src/analysis/has_public_example.d
+++ b/src/analysis/has_public_example.d
@@ -1,0 +1,302 @@
+// Distributed under the Boost Software License, Version 1.0.
+//	  (See accompanying file LICENSE_1_0.txt or copy at
+//			http://www.boost.org/LICENSE_1_0.txt)
+
+module analysis.has_public_example;
+
+import analysis.base;
+import dsymbol.scope_ : Scope;
+import dparse.ast;
+import dparse.lexer;
+
+import std.algorithm;
+import std.stdio;
+
+/**
+ * Checks for public declarations without a documented unittests.
+ * For now, variable and enum declarations aren't checked.
+ */
+class HasPublicExampleCheck : BaseAnalyzer
+{
+	alias visit = BaseAnalyzer.visit;
+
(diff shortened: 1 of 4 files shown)


src/analysis/has_public_example.d
```diff
@@ -0,0 +1,234 @@
...
+        Declaration lastDecl;
+        bool hasPublicUnittest = true; // reset on lastDecl assignment
+
+		// only look at top-level declarations
+        foreach (decl; mod.declarations)
+        {
+            if (!isPublic(decl.attributes))
+                continue;
+
+            bool hasDdocHeader = hasDdocHeader(decl);
```
BBasile: `const bool`

src/analysis/has_public_example.d
```diff
@@ -0,0 +1,234 @@
...
+
+	this(string fileName, const(Scope)* sc, bool skipTests = false)
+	{
+		super(fileName, sc, skipTests);
+	}
+
+    override void visit(const Module mod)
+    {
+        Declaration lastDecl;
+        bool hasPublicUnittest = true; // reset on lastDecl assignment
```
BBasile: I'd rather name this variable `hasNoPublicTest` and change the related relational expressions, because

- then you'll can remove the initializer.
- if you look at the use spots, you always negate the variable.

hence you'll have the same meaning with less complexity.

wilzbach:
> I'd rather name this variable `hasNoPublicTest` and change the related relational expressions, because

Excellent idea!

wilzbach:
Merging this as this is a new, by-default disabled check & I can "battle-test" it on Phobos over the next weeks.

PR #475, opened 2017-06-22
Fixes cases of false or non positive with the useless init check

### LINKED ISSUE #474: Case of false positive with the useless init check : manifest constants
```d
enum byte a;
```

For this case it's even an error not to put an initializer.

BBasile:
This PR fixes 4 bugs found while applying the useless initializer check to phobos.

### DIFF
--- a/src/analysis/base.d
+++ b/src/analysis/base.d
@@ -56,7 +56,7 @@ public:
 
 protected:
 
-	bool inAggregate = false;
+	bool inAggregate;
     bool skipTests;
 
 	template visitTemplate(T)
--- a/src/analysis/useless_initializer.d
+++ b/src/analysis/useless_initializer.d
@@ -15,9 +15,9 @@ import std.stdio;
 
 /*
 Limitations:
-    - Stuff s = Stuff.init does not work with type with postfixes`*` `[]`.
-    - Stuff s = Stuff.init is only detected for struct within the module.
-    - BasicType b = BasicType(v), default init used in BasicType ctor, e.g int(8).
+	- Stuff s = Stuff.init does not work with type with postfixes`*` `[]`.
+	- Stuff s = Stuff.init is only detected for struct within the module.
+	- BasicType b = BasicType(v), default init used in BasicType ctor, e.g int(8).
 */
 
 /**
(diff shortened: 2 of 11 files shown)


BBasile:
Right now with these fixes i just have **1** warning in phobos.

> ./std/conv.d(4385:28)[warn]: Variable init initializer is useless because it does not differ from the default value

BBasile:
This is good now. After the PR mentioned above and if you run
`dscanner -S | grep "initializer is useless" > dscanner.txt` in phobos repository then a 0 Kio file is produced.

src/analysis/useless_initializer.d
```diff
@@ -25,45 +32,128 @@ final class UselessInitializerChecker : BaseAnalyzer
...
+    }
 
     static immutable strDefs = [`""`, `""c`, `""w`, `""d`, "``", "``c", "``w", "``d", "q{}"];
     static immutable intDefs = ["0", "0L", "0UL", "0uL", "0U", "0x0", "0b0"];
 
+    HashMap!(string, bool) _structCanBeInit;
+    DynamicArray!(string) _structStack;
+    DynamicArray!(bool) _inStruct;
+    DynamicArray!(bool) _atDisabled;
+    DynamicArray!(bool) _inTest;
```
wilzbach: Hmm I am not a huge fan of using `_` if there's a proper type system with `private` in place.

src/analysis/useless_initializer.d
```diff
@@ -25,45 +32,128 @@ final class UselessInitializerChecker : BaseAnalyzer
...
+    override void visit(const(Unittest) test)
+    {
+        _inTest.insert(true);
+        test.accept(this);
+        _inTest.removeBack();
+    }
+
+    override void visit(const(StructDeclaration) decl)
+    {
+        if (_inTest.back())
```
wilzbach: If you always just test the end of the array, wouldn't a simple `bool` flag suffice?

src/analysis/useless_initializer.d
```diff
@@ -25,45 +32,128 @@ final class UselessInitializerChecker : BaseAnalyzer
...
+        }
     }
 
     override void visit(const(VariableDeclaration) decl)
     {
+        if (!decl.type || !decl.type.type2 ||
+            // initializer has to appear clearly in generated ddoc
+            decl.comment !is null ||
+            // issue 474, manifest constants HAVE to be initialized.
+            !decl.storageClasses.filter!(a => a.token == tok!"enum").empty)
```
wilzbach: How about using `canFind`?

src/analysis/useless_initializer.d
```diff
@@ -5,13 +5,20 @@
 module analysis.useless_initializer;
 
 import analysis.base;
+import containers.dynamicarray;
+import containers.hashmap;
 import dparse.ast;
 import dparse.lexer;
+import std.algorithm : among, canFind;
+import std.algorithm.iteration : filter;
```
wilzbach: Why not simply importing entire `std.algorithm` - it's very commonly used... and saves the pain of updating this list all the time.

src/analysis/useless_initializer.d
```diff
@@ -25,45 +32,128 @@ final class UselessInitializerChecker : BaseAnalyzer
...
+            if (!declarator.initializer || !declarator.initializer.nonVoidInitializer ||
+                declarator.comment !is null)
+                    continue;
 
             version(unittest)
+            {
                 enum warn = q{addErrorMessage(declarator.name.line, declarator.name.column,
                     key, msg);};
+            }
             else
```
wilzbach: Off-topic: I am not a huge friend of this pattern. While I do the see merits of being able to just use `X`, the string formatting could (in theory of course) be broken.

src/analysis/useless_initializer.d
```diff
@@ -25,45 +32,128 @@ final class UselessInitializerChecker : BaseAnalyzer
...
+        }
+        _inStruct.removeBack();
+    }
+
+    override void visit(const(Constructor) decl)
+    {
+        if (_inStruct.length > 1 && _inStruct[$-2] &&
+            ((decl.parameters && decl.parameters.parameters.length == 0) || !decl.parameters))
+        {
+            _structCanBeInit[_structStack.back()] = !_atDisabled[$-1];
```
wilzbach: While it shouldn't matter, we could save the duplicate hashmap query here:

```
bool canBeInit = !_atDisabled[$-1];
if ...

_structCanBeInit[_structStack.back()] = canBeInit
```

src/analysis/useless_initializer.d
```diff
@@ -25,45 +32,128 @@ final class UselessInitializerChecker : BaseAnalyzer
...
     else
+    {
         enum msg = `Variable %s initializer is useless because it does not differ from the default value`;
+    }
 
     static immutable strDefs = [`""`, `""c`, `""w`, `""d`, "``", "``c", "``w", "``d", "q{}"];
     static immutable intDefs = ["0", "0L", "0UL", "0uL", "0U", "0x0", "0b0"];
 
+    HashMap!(string, bool) _structCanBeInit;
+    DynamicArray!(string) _structStack;
```
wilzbach: As mentioned below: I think `_structStack`, `_atDisabled` and `_inTest` don't need to be arrays.

src/analysis/useless_initializer.d
```diff
@@ -25,45 +32,128 @@ final class UselessInitializerChecker : BaseAnalyzer
...
+        test.accept(this);
+        _inTest.removeBack();
+    }
+
+    override void visit(const(StructDeclaration) decl)
+    {
+        if (_inTest.back())
+            return;
+
+        _structStack.insert(decl.name.text);
```
wilzbach: Hmm are you sure this would be unique? Wouldn't you need proper mangling?
Consider:

```d
struct F {
	struct G {}
	G g;
}
struct G {
	struct G {} // collision
	G g;
}
```

src/analysis/useless_initializer.d
BBasile: because of this case:

```d
unittest{struct A{unittest{}}}
```

src/analysis/useless_initializer.d
BBasile: I'll use the FQN

src/analysis/base.d
```diff
@@ -56,7 +56,7 @@ public:
 
 protected:
 
-	bool inAggregate = false;
+	bool inAggregate;
```
skl131313: Looks like there's a tab here instead of spaces, even before the change.

BBasile: For sure there's one. @Hackerpilot uses tabs on his stuff. Excepted libdparse because it was planned to be included in phobos.

Hackerpilot: The `.editorconfig` file for this project does specify tabs.

src/analysis/useless_initializer.d
```diff
@@ -25,49 +31,142 @@ final class UselessInitializerChecker : BaseAnalyzer
...
+    {
+        _inStruct.insert(decl.structDeclaration !is null);
+        decl.accept(this);
+        if (_inStruct.length > 1 && _inStruct[$-2] && decl.constructor &&
+            ((decl.constructor.parameters && decl.constructor.parameters.parameters.length == 0) ||
+            !decl.constructor.parameters))
+        {
+            _atDisabled[$-1] = decl.attributes
+                .canFind!(a => a.atAttribute !is null && a.atAttribute.identifier.text == "disable");
+        }
```
wilzbach: OT: At some point we probably should start sharing code between the checkers. I imagine sth. like `isInDisabledStruct` to be a fairly common request.

src/analysis/useless_initializer.d
```diff
@@ -25,49 +31,143 @@ final class UselessInitializerChecker : BaseAnalyzer
...
+            // initializer has to appear clearly in generated ddoc
+            decl.comment !is null ||
+            // issue 474, manifest constants HAVE to be initialized.
+            decl.storageClasses.canFind!(a => a.token == tok!"enum"))
         {
-            if (!decl.type || !decl.type.type2)
-                continue;
-            if (!declarator.initializer || !declarator.initializer.nonVoidInitializer)
-                continue;
+            return;
```
wilzbach: Speaking of tabs, this looks like whitespace here. Should we maybe enable a checker on Travis, s.t. we can't push files which are beginning with tabs?

wilzbach: (Sorry that I didn't realize this earlier - I don't care about the holy war of tabs vs. whitespace, but we should try to keep this repo consistent)

BBasile: I'm gonna patch this. I still have to add editorconfig support in Coedit. Currently the tab style  is autodetected and for a new module the user preference is used, so spaces here.

wilzbach:
> I'm gonna patch this. I still have to add editorconfig support in Coedit. Currently the tab style is autodetected and for a new module the user preference is used, so spaces here.

Ok. The history got moderately long for this diff. Do you want to squash everything before merging?
(I don't mind either way).
Btw should we define a guideline for this at https://github.com/dlang-community/discussions?

BBasile:
Yes squash. If GH wouldn't allow to do it i would have rebased and marked `fixup` all the commits.

PR #479, opened 2017-06-25
Run DScanner on itself

wilzbach:


src/analysis/imports_sortedness.d
```diff
@@ -66,7 +66,7 @@ class ImportSortednessCheck : BaseAnalyzer
 
 private:
 
-	int level = 0;
```
wilzbach: FWIW I'm not really sold on the useless initializer check, but there weren't that many occurrences.

src/analysis/style.d
```diff
@@ -20,9 +20,9 @@ final class StyleChecker : BaseAnalyzer
 {
 	alias visit = ASTVisitor.visit;
 
-	enum string varFunNameRegex = `^([\p{Ll}_][_\w\d]*|[\p{Lu}\d_]+)$`;
-	enum string aggregateNameRegex = `^\p{Lu}[\w\d]*$`;
-	enum string moduleNameRegex = `^[\p{Ll}_\d]+$`;
+	enum string VarFunNameRegex = `^([\p{Ll}_][_\w\d]*|[\p{Lu}\d_]+)$`;
+	enum string AggregateNameRegex = `^\p{Lu}[\w\d]*$`;
+	enum string ModuleNameRegex = `^[\p{Ll}_\d]+$`;
```
wilzbach: It's a bit ironic if a checker violates its own checks ;-)

src/analysis/if_statements.d
```diff
@@ -117,6 +117,17 @@ private struct ExpressionInfo
 			return 0;
 		return -1;
 	}
+	bool opEquals(ref const ExpressionInfo other) const nothrow
+	{
+		return formatted == other.formatted
+			&& line == other.line
+			&& column == other.column
+			&& depth == other.depth;
```
ZombineDev: Doesn't this contradict the hash function, because it takes the `formatted` into account?

src/analysis/if_statements.d
```diff
@@ -117,6 +117,17 @@ private struct ExpressionInfo
...
+	}
+	size_t toHash() const nothrow
+	{
+		return line * 100_000 + column * 100 + depth;
```
ZombineDev: I think better to rely on the built-in hashing here: https://github.com/dlang/druntime/blob/master/src/core/internal/hash.d

src/analysis/if_statements.d
```diff
@@ -117,6 +117,17 @@ private struct ExpressionInfo
 			return 0;
 		return -1;
 	}
+	bool opEquals(ref const ExpressionInfo other) const nothrow
```
ZombineDev: Actually, are these functions added because of `object_const_check`? If that's the case I think this is a false positive, since the compiler already generates those functions.

src/analysis/imports_sortedness.d
BBasile: Speaking of that i still have a PR opened that make the check very reliable. (#475)

src/analysis/pokemon.d
```diff
@@ -67,8 +67,7 @@ class PokemonExceptionCheck : BaseAnalyzer
 		{
 			return;
 		}
-		auto identOrTemplate = type2.symbol.identifierOrTemplateChain
-			.identifiersOrTemplateInstances[0];
+		const identOrTemplate = type2.symbol.identifierOrTemplateChain.identifiersOrTemplateInstances[0];
 		if (identOrTemplate.templateInstance !is null)
```
ZombineDev: Um, why? This lines is already 106 colons with the two 4-width tabs.

src/analysis/imports_sortedness.d
ZombineDev: TBH, despite the recent discussion, I find code without explicit initializers (when equal to .init) more clear to read, so LGTM.

src/analysis/style.d
ZombineDev: No, manifest constants like `enum [type] identifier = <init-expr>;`are variables, while `enum`s like `enum EnumerationName { member1, meber2 }` are aggregates, IMO.

src/analysis/imports_sortedness.d
ZombineDev: @BBasile I'll try to review your PR, but I don't have much knowledge of D-Scanner's internals.

src/analysis/unmodified.d
```diff
@@ -198,7 +198,7 @@ private:
 		auto vi = VariableInfo(name);
 		if (guaranteeUse == 0)
 		{
-			const r = tree[index].equalRange(&vi);
+			auto r = tree[index].equalRange(&vi);
```
wilzbach: This lead to the compile error on Travis:

> src/analysis/unmodified.d(202): Error: mutable method std.container.rbtree.RBRange!(RBNode!(VariableInfo*)*).RBRange.front is not callable using a const object

wilzbach:
@ZombineDev 

> Actually, are these functions added because of object_const_check?

It's due to the `opequals_tohash_check`, but I think it's better to go with a minimal load of checks in the beginning and discuss on a solution for enabling these checks on a case by case basis ;-)
That's why I removed the controversial bits and set the checks to disabled.
Of course, we should squash all commits into one.

ZombineDev:
> It's due to the opequals_tohash_check

Yeah, I was looking for `opequals_tohash_check`, but somehow I only found `object_const_check` when searching for toHash. Strange.

> That's why I removed the controversial bits and set the checks to disabled.

Yes, we can always discus them later.

src/analysis/unmodified.d
ZombineDev: I'll submit a Phobos PR for that (I already have the fix locally, but I want to add some tests), but that's the right way to go, since we support multiple versions of Phobos anyway.

src/analysis/style.d
ZombineDev: ping, this is the only issue left ;)

ZombineDev:
> Of course, we should squash all commits into one.

Yep, no problem, but individual commits are extremely helpful when reviewing ;)

makefile
```diff
@@ -51,6 +51,9 @@ test: githash
 	./bin/dscanner-unittest
 	rm -f bin/dscanner-unittest
 
+lint: dmdbuild
+	./dsc --config .dscanner.ini --styleCheck src
```
ZombineDev: `./dsc --config .dscanner.ini --styleCheck src` -> 
`./bin/dscanner --config .dscanner.ini --styleCheck src`

.travis.sh
```diff
@@ -10,4 +10,5 @@ elif [[ $DC == ldc2 ]]; then
 else
     git submodule update --init --recursive
     make test
+    make lint
```
ZombineDev: I heavily prefer if we ran `lint` on both the ldc, dmd and the dub build.

wilzbach:
> I heavily prefer if we ran `lint` on both the ldc, dmd and the dub build.

Hmm I don't mind, but on my machine it takes ~40s (~1.5min on CircleCI) to do the release build and the source code doesn't change ...
Btw if we already talk about the Makefile I would be strongly in favor of
- allowing a "slim" debug build (=debug without logging, at Phobos we already patch DScanner for this as building the debug version is _a lot_ faster)
- use non-changing target that don't run when no source file has changed (with the current makefile `make lint` will always build DScanner)

ZombineDev:
> Hmm I don't mind, but on my machine it takes ~40s (~1.5min on CircleCI) to do the release build and the source code doesn't change ...

The idea is more about testing the compilers than anything else, but I'm fine with leaving it as it is.

ZombineDev:
I decided to move compiler testing discussion to different PR/issue, as the core of the PR was complete.

PR #484, opened 2017-06-28
Check for function parameters in mixins

wilzbach:
This isn't a proper fix for the unused variable errors in has_public_example, but me publicly experimenting with it.

@BBasile Is there a particular reason why you still trigger error messages (even if the message is detected as "uncertain"?)

### DIFF
--- a/src/dscanner/analysis/unused.d
+++ b/src/dscanner/analysis/unused.d
@@ -412,6 +412,8 @@ private:
 		{
 			if (!uu.isRef && tree.length > 1)
 			{
+			    if (uu.uncertain)
+			        continue;
 				immutable string certainty = uu.uncertain ? " might not be used."
 					: " is never used.";
 				immutable string errorMessage = (uu.isParameter ? "Parameter " : "Variable ")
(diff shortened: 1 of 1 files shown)


BBasile:
Can you rebase for inclusion in 0.5.0 ?

src/dscanner/analysis/unused.d
```diff
@@ -407,6 +412,8 @@ private:
 		{
 			if (!uu.isRef && tree.length > 1)
 			{
+			    if (uu.uncertain)
+			        continue;
 				immutable string certainty = uu.uncertain ? " might not be used."
 					: " is never used.";
```
wilzbach: This line would need adapting as well if this is considered for inclusion.

BBasile: Sorry but i don't see the problem, can you explain more ?

wilzbach: If we go with this change this line should be rewritten to `immutable string certainty = " is never used."; as "might" now no longer can happen.
Though to be honest it's quite a long time ago since I made this change.

PR #491, opened 2017-07-04
output in bin when using the bat and DUB

BBasile:
For consistency with the makefile.


BBasile:
the json has to be updated too

PR #495, opened 2017-07-08
Add check for asserts without an explanatory message

wilzbach:
From the NG: http://forum.dlang.org/post/uagjlugnczfegbuxdjzr@forum.dlang.org

### DIFF
--- a/src/analysis/assert_without_msg.d
+++ b/src/analysis/assert_without_msg.d
@@ -10,6 +10,7 @@ import dparse.lexer;
 import dparse.ast;
 
 import std.stdio;
+import std.algorithm;
 
 /**
  * Check that all asserts have an explanatory message.
@@ -31,8 +32,60 @@ class AssertWithoutMessageCheck : BaseAnalyzer
 			addErrorMessage(expr.line, expr.column, KEY, MESSAGE);
 	}
 
+	override void visit(const FunctionCallExpression expr)
+	{
+		if (!isStdExceptionImported)
+			return;
+
+		if (expr.unaryExpression !is null &&
+			expr.unaryExpression.primaryExpression !is null &&
+			expr.unaryExpression.primaryExpression.identifierOrTemplateInstance !is null)
+		{
+			auto ident = expr.unaryExpression.primaryExpression.identifierOrTemplateInstance.identifier;
+			if (ident.text == "enforce" && expr.arguments !is null && expr.arguments.argumentList !is null &&
+					expr.arguments.argumentList.items.length < 2)
+				addErrorMessage(ident.line, ident.column, KEY, MESSAGE);
+		}
+	}
+
+	override void visit(const SingleImport sImport)
+	{
+		static immutable stdException = ["std", "exception"];
+		if (sImport.identifierChain.identifiers.map!(a => a.text).equal(stdException))
+			isStdExceptionImported = true;
+	}
+
+	// revert the stack after new scopes
+	override void visit(const Declaration decl)
+	{
+		// be careful - ImportDeclarations don't introduce a new scope
+		if (decl.importDeclaration is null)
+		{
+			bool tmp = isStdExceptionImported;
+			scope(exit) isStdExceptionImported = tmp;
+			decl.accept(this);
+		}
+		else
+			decl.accept(this);
+	}
+
+	mixin ScopedVisit!IfStatement;
+	mixin ScopedVisit!BlockStatement;
+
 	alias visit = BaseAnalyzer.visit;
 
+private:
+	bool isStdExceptionImported;
+
+	template ScopedVisit(NodeType)
+	{
+		override void visit(const NodeType n)
+		{
+			bool tmp = isStdExceptionImported;
+			scope(exit) isStdExceptionImported = tmp;
+			n.accept(this);
+		}
+	}
 }
 
 unittest
@@ -63,6 +116,40 @@ unittest
 		AssertWithoutMessageCheck.MESSAGE,
 	), sac);
 
+	// check for std.exception.enforce
+	assertAnalyzerWarnings(q{
+	unittest {
+		enforce(0); // std.exception not imported yet - could be a user-defined symbol
+		import std.exception;
+		enforce(0, "foo bar");
+		enforce(0); // [warn]: %s
+	}
+	}c.format(
+		AssertWithoutMessageCheck.MESSAGE,
+	), sac);
+
+	// check for std.exception.enforce
+	assertAnalyzerWarnings(q{
+	unittest {
+		import exception;
+		class C {
+			import std.exception;
+		}
+		enforce(0); // std.exception not imported yet - could be a user-defined symbol
+		struct S {
+			import std.exception;
+		}
+		enforce(0); // std.exception not imported yet - could be a user-defined symbol
+		if (false) {
+			import std.exception;
+		}
+		enforce(0); // std.exception not imported yet - could be a user-defined symbol
+		{
+			import std.exception;
+		}
+		enforce(0); // std.exception not imported yet - could be a user-defined symbol
+	}
+	}c, sac);
 
 	stderr.writeln("Unittest for AssertWithoutMessageCheck passed.");
 }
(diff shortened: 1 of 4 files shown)


CyberShadow:
Awesome! `enforce` too, please?

ZombineDev:
I agree with @BBasile that `assert(0)` and `assert(false)` could be left out.

ZombineDev:
But in general, when reasonable people may disagree, it's better to leave it to the end-user to decide. So I guess the solution should be to add `assert0_without_msg` in addition to `assert_without_msg` and analogically for enforce.

Hopefully there should be an easy way to reuse the same analyzer for all of them. I wonder if something along the lines of this would work:

```D
enum AssertCheck
{
    normalAssert,
    unreachableAssert,
    normalEnforce,
    unreachableEnforce
}

class AssertWithoutMessageCheck : AssertWithoutMessageCheckImpl!(AssertCheck.normalAssert) { }
class Assert0WithoutMessageCheck : AssertWithoutMessageCheckImpl!(AssertCheck.unreachableAssert) { }
class EnforceWithoutMessageCheck : AssertWithoutMessageCheckImpl!(AssertCheck.normalEnforce) { }
class Enforce0WithoutMessageCheck : AssertWithoutMessageCheckImpl!(AssertCheck.unreachableEnforce) { }

class AssertWithoutMessageCheckImpl(AssertCheck config) : BaseAnalyzer
{
    static if (config == AssertCheck.normalAssert)
    {
        enum string KEY = "dscanner.style.assert_without_msg";
        enum string MESSAGE = "An assert(cond) should have an explanatory message";
    }
    else static if (config == AssertCheck.unreachableAssert)
    {
        enum string KEY = "dscanner.style.assert0_without_msg";
        enum string MESSAGE = "An assert(false) should have an explanatory message";
    }
    else static if (config == AssertCheck.normalEnforce)
    {
        enum string KEY = "dscanner.style.enforce_without_msg";
        enum string MESSAGE = "An enforce(cond) should have an explanatory message";
    }
    else static if (config == AssertCheck.unreachableEnforce)
    {
        enum string KEY = "dscanner.style.enforce0_without_msg";
        enum string MESSAGE = "An enforce(false) should have an explanatory message";
    }
    else
        static assert(0, "Invalid AssertCheck template argument: " ~ config.stringof);

    // ...
}
```

Sounds good?

BBasile:
@wilzbach, did you try the new check with phobos ? Maybe the results, the false positive and how much solvable they look can help to convince you but it's only two cases to skip, 0 and false and empty string literals too.

src/analysis/config.d
```diff
@@ -190,7 +190,7 @@ struct StaticAnalysisConfig
 	string has_public_example = Check.disabled;
 
 	@INI("Check for asserts without an explanatory message")
-	string assert_without_msg = Check.disabled;
+	string assert_without_msg = Check.skipTests;
```
BBasile: I don't agree with this change. Wasn't the check designed for phobos in the first place ?

BBasile: Meaning that disabled was okay.

wilzbach:
> Awesome! enforce too, please?

Done.

> @wilzbach, did you try the new check with phobos ? Maybe the results, the false positive and how much solvable they look can help to convince you but it's only two cases to skip, 0 and false and empty string literals too.

Good idea. Full list is here: http://sprunge.us/MaGf

A couple of examples (abbreviated):

### `assert(false)`

From `std.functional`:

```d
private uint _ctfeSkipOp(ref string op)
{
    if (!__ctfe) assert(false);
...
```

### `assert(0)`

From `std.math`:

```d
        else version (Solaris)
        {
            asm pure nothrow @nogc // assembler by W. Bright
            {
                // EDX = (A.length - 1) * real.sizeof
...
            }
        }
        else
        {
            static assert(0);
        }
```

One from `std.process`:

```d
    void remove(in char[] name) @trusted nothrow @nogc // TODO: @safe
    {
        version (Windows)    SetEnvironmentVariableW(name.tempCStringW(), null);
        else version (Posix) core.sys.posix.stdlib.unsetenv(name.tempCString());
        else static assert(0);
    }
```

From `std.socket`:

```d
else
{
    static assert(0);     // No socket support yet.
}
```

Imho it would be better to have e.g. the message "No socket support yet" directly appear on the CLI, hence I don't see `assert(0)` as a false positive.

> Also it's very easy to cheat on the check: assert(0, " ")`.

At least for peer-reviewed library, these cheats will/should be noticed.
I added a simple "cheat prevention", so "   " will be detected .

> I don't agree with this change. Wasn't the check designed for phobos in the first place ?

Fair enough.

wilzbach:
> Hopefully there should be an easy way to reuse the same analyzer for all of them.

I think there's no usecase to prevent `enforce(0)` and allow `assert(0)`, so we could simplify it to to checks.

> I wonder if something along the lines of this would work:

DScanner isn't used as an API, but through the command-line and thus the `dscanner.ini` is the _only_ way to customize the behavior. There's already a default string defined for all tests:

```d
enum Check: string
{
    /// Check is disabled.
    disabled    = "disabled",
    /// Check is enabled.
    enabled     = "enabled",
    /// Check is enabled but not operated in the unittests.
    skipTests   = "skip-unittest"
}
```

But we could create a new type for this check - sth. like this would work fine:

```d
enum AssertWithoutMsgCheck: string
{
    /// Check is disabled.
    disabled    = "disabled",
    /// Check is enabled.
    enabled     = "enabled",
    /// Check is enabled but not operated in the unittests.
    skipTests   = "skip-unittest",
   /// Check is enabled, but not for assert(0) or assert(false)
   skipAssert0   = "skip-assert0",
   /// Check is enabled, but not for assert(0) or assert(false) and unittest
   skipTestsAssert0   = "skip-unittest-assert0",
}
```

src/analysis/assert_without_msg.d
```diff
@@ -0,0 +1,186 @@
...
+	{
+		override void visit(const NodeType n)
+		{
+			bool tmp = isStdExceptionImported;
+			scope(exit) isStdExceptionImported = tmp;
+			n.accept(this);
+		}
+	}
+
+	void preventCheating(const ExpressionNode exprNode)
```
CyberShadow: IMHO this is a waste of code.

wilzbach: I agree, but it wasn't hard and it helps to get this through the review.

CyberShadow: The cheating complaint is nonsensical, someone inclined to "cheat" can just as well do it with `assert(cond, "fgsfds");`. There is no need to waste any time on it.

BBasile: which you would also detect while reviewing Phobos, if this assert thing was really an issue (caugh caugh).

CyberShadow: It's easy to write an assertion or enforce without a condition by accident, negligence, or simple unawareness of best practices; however, cheating the tool would take malevolent intent. We pretty much only care about the first three.

BBasile:
I still dont see the code that test for `assert(0)` or `assert(false)`.

CyberShadow:
> I still dont see the code that test for assert(0) or assert(false).

It seems like we don't need that for Phobos.

If you'd like it for general use of DScanner, perhaps you could implement that after this PR is merged?

src/analysis/assert_without_msg.d
```diff
@@ -0,0 +1,186 @@
...
+	import analysis.config : StaticAnalysisConfig, Check, disabledConfig;
+	import analysis.helpers : assertAnalyzerWarnings;
+
+	StaticAnalysisConfig sac = disabledConfig();
+	sac.assert_without_msg = Check.enabled;
+
+	assertAnalyzerWarnings(q{
+	unittest {
+		assert(0, "foo bar");
+		assert(0); // [warn]: %s
```
wilzbach: @BBasile FWIW I test for `assert(0)` - just not in the way you would like it to be.

BBasile:
I opt out from this review. The only thing that's important to my eyes is that it's disabled by default.

BBasile:
Sorry, hit the wrong button.

src/analysis/assert_without_msg.d
Hackerpilot: I agree with @CyberShadow, we should protect against accidents, not malice.

BBasile:
So nobody merges this ? Was it supposed to be finished or WIP ?

wilzbach:
> Was it supposed to be finished or WIP ?

I never had time to remove the "cheating check", but AFAICT nothing else is blocking this.

BBasile:
So removes it and let's get the merge.

wilzbach:
> So removes it and let's get the merge.

Finally removed!

PR #502, opened 2017-07-25
Fix #501

### LINKED ISSUE #501: False default initializer when initializing with ""
```d
string a;
writeln(a is null); // true
string b = ""; // initializer is useless
writeln(b is null); // false
string c = null; // nothing
writeln(c is null); // true
```
As seen here `a` is null by default ant not `""`, so the initializer warning is showing wrong information

WebFreak001:
because `"" !is string.init` and is sometimes what you want to return of a function instead of `null`

### DIFF
--- a/src/analysis/useless_initializer.d
+++ b/src/analysis/useless_initializer.d
@@ -41,7 +41,6 @@ private:
 		enum msg = `Variable %s initializer is useless because it does not differ from the default value`;
 	}
 
-	static immutable strDefs = [`""`, `""c`, `""w`, `""d`, "``", "``c", "``w", "``d", "q{}"];
 	static immutable intDefs = ["0", "0L", "0UL", "0uL", "0U", "0x0", "0b0"];
 
 	HashMap!(string, bool) _structCanBeInit;
@@ -217,7 +216,7 @@ public:
 					if (intDefs.canFind(value.text))
 						mixin(warn);
 				}
-				else if (isPtr)
+				else if (isPtr || isStr)
 				{
 					if (str(value.type) == "null")
 						mixin(warn);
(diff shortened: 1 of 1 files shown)


BBasile:
I disagree. `string a = "";` is clearly a useless initializer.

WebFreak001:
```d
/// Returns: null if unsuccessful
string func()
{
  string ret = ""; // annoying wrong warning
  if (error)
    ret = null;
  return ret;
}
```

It does make sense, many string functions differentiate between `""` and `null` when returning. I don't want to get spammed with so many wrong dscanner warnings like this.

BBasile:
The rule of thumb with D arrays is  "what counts is the `.length` property". All the cases that are handled by this check are compliant with this rule. The code you show is a really ugly abuse of D arrays to return a kind of error code.

WebFreak001:
This is just example code, an actual implementation could receive from some TCP stream instead and instead of allocating a char[] for a result there return null which might even happen in two steps (making a local variable and then returning it)

WebFreak001:
Or here I have an example from actual code:
```d
struct Part
{
  string contentType = null; /// Set to a string to clear or replace the content type, leave null to inherit
}
```

BBasile:
Use the config file to disable the check.

WebFreak001:
the problem is that dscanner tells me and other users wrong information, everyone will assume the initializer is not null if they use `""`

BBasile:
Again i think you abuse of an implementation detail.
https://forum.dlang.org/post/mailman.2307.1382052843.1719.digitalmars-d@puremagic.com.
Even the language architect disagree with the usefulness of making the difference:
https://forum.dlang.org/post/l3rktv$uj9$1@digitalmars.com

BBasile:
I don't like to split the community so you'll get merged. However i think that other people's mind would have been necessary to solve this controversy point of view.

BBasile:
That being said i still think i was right. The point is that it's not that important, it's just a stupid and blind static check. The cases it covers shall be sufficient to make people understand that they don't need to initialize their variables..._"in case of..."_

PR #503, opened 2017-07-27
fix deprecations messages related to message function

BBasile:
Since three days libdparse prefers a delegate.


BBasile:
sorry to self-merge but there's not much to discuss there. Soon or later this would had to be done.

PR #508, opened 2017-08-03
final attrib checker - handle static

BBasile:
Static methods are never virtual therefore the final attribute is superfluous. Found a case in [Netflix vectorFlow ;)](https://github.com/Netflix/vectorflow/blob/master/src/vectorflow/optimizers.d#L353).


BBasile:
Is this good ? I want to merge because i wrote this check but well...self-merge is against the community rule.

wilzbach:
> Is this good ? 

Yeah I definitely agree on it being a useful addition, but I'm really low on time at the moment. Will try to have a look at it tonight. Sorry

BBasile:
No problem i just need an agreement for the form. The check already exists and for static I test the 3 ways for which it can declared (function - colon - block).

PR #509, opened 2017-08-05
fix #370 - False positive for duplicate variable name check with structs

### LINKED ISSUE #370: False positive for duplicate variable name check with structs
``` d
int a;
struct S
{
    int a;
}
```

BBasile:


ZombineDev:
Can you add a test for:
```D
int aa;
struct a { int a; }
```
?

BBasile:
The fqn is joined with a period so it should be okay a priori.

BBasile:
As expected it passes. I think you should have you miss the `join(".")`.

PR #510, opened 2017-08-06
fix #452 - false pos for naming style of a VariableDecl with the "enum" storage class

### LINKED ISSUE #452: false positive for naming style with the enum storage class
```d
enum bool aZ = true;
```

> [warn]: Variable name 'aZ' does not match style guidelines.

BBasile:
This might be debatable but the issue is opened for weeks and was never flagged as invalid.


rikkimax:
It's been almost a week.
Supposedly it works fine (according to @BBasile).

I'm pulling.

PR #512, opened 2017-08-12
Add a findDeclarationOf overload that takes a callback

WebFreak001:
This makes getting the declarations when working with dscanner as a library much easier. Another possible improvement for the future could be directly passing File objects or an input range of Files.

Also this removed a bit of code duplication where there were so many writefln calls.

### DIFF
--- a/src/symbol_finder.d
+++ b/src/symbol_finder.d
@@ -16,7 +16,8 @@ import std.functional : toDelegate;
 
 void findDeclarationOf(File output, string symbolName, string[] fileNames)
 {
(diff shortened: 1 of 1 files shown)


src/symbol_finder.d
```diff
@@ -16,6 +16,21 @@ import std.functional : toDelegate;
 
 void findDeclarationOf(File output, string symbolName, string[] fileNames)
 {
+	findDeclarationOf((string fileName, size_t line, size_t column) {
```
BBasile: Hallman curly braces plz

PR #519, opened 2017-08-24
update dparse - could stuck because of AA lit bug

BBasile:
A bug that's fixed in 0.7.1 could make D-Scanner falling into an infinite loop, see https://github.com/dlang-community/libdparse/issues/167


BBasile:
I need this asap because i release a Coedit update, which include the pre-built D-Scanner.

PR #521, opened 2017-09-11
Upgrade libdparse for static foreach [phobos]

wilzbach:


wilzbach:
Merging to `phobos`, s.t. we can propagate this ASAP and it's a simple version bump anyhow.

PR #523, opened 2017-10-01
Compile DScanner with -fPIC

wilzbach:
[By default, DMD passes `-fPIC` to the linker](https://github.com/dlang/dmd/blob/master/ini/linux/bin64/dmd.conf), but sadly the development build of DMD doesn't: https://github.com/dlang/dmd/pull/7002

As the regarding PR won't be merged in the foreseeable future, I think it's best to simply add the flag to the Makefile.


wilzbach:
No one objected for six weeks and it's a trivial Makefile improvement -> merging.

PR #525, opened 2017-10-10
apply dfmt

stefan-koch-sociomantic:
I've just applied dfmt such that my editor stops complaining.
Maybe it's worth a pull ?

### DIFF
--- a/src/analysis/allman.d
+++ b/src/analysis/allman.d
@@ -34,14 +34,14 @@ class AllManCheck : BaseAnalyzer
 		foreach (i; 1 .. tokens.length - 1)
 		{
 			const curLine = tokens[i].line;
-			const prevTokenLine = tokens[i-1].line;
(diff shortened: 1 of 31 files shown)


stefan-koch-sociomantic:
ldc on travis fails b/c of network timeouts.

wilzbach:
While I understand the motivation, this will cause me a lot of pain (rebasing all my PRs and branches) with almost no gain.
However, if there's consensus to go this way, I wouldn't object.
CC @Hackerpilot @BBasile @ZombineDev

stefan-koch-sociomantic:
@wilzbach well apply dfmt to your prs.
it should result in the same diff

wilzbach:
Fair enough, but git blame will be affected too.

BBasile:
Two args for closing.

- DFMT is not mature enough
- unmaintained.

PR #534, opened 2017-11-19
fix #533 - "Could be declared const" should be avoided for declarations with "new"

### LINKED ISSUE #533: "Could be declared const" should be avoided for declarations with "new"
In D, `new` is mostly used to create classes. D-Scanner warns for

```d
E e = new E;
```

that _"Variable e is never modified and could have been declared const or immutable"_ but because of `const` transitivity we rarely want the instance to be `const`.

BBasile:
see #533 for rationale.

### DIFF
--- a/src/analysis/unmodified.d
+++ b/src/analysis/unmodified.d
@@ -58,6 +58,8 @@ class UnmodifiedFinder : BaseAnalyzer
 			{
 				if (initializedFromCast(d.initializer))
 					continue;
+				if (initializedFromNew(d.initializer))
+					continue;
 				tree[$ - 1].insert(new VariableInfo(d.name.text, d.name.line,
 						d.name.column, isValueTypeSimple(dec.type)));
 			}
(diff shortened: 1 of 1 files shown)


src/analysis/unmodified.d
```diff
@@ -211,6 +215,19 @@ private:
 		}
 	}
 
+	bool initializedFromNew(const Initializer initializer)
+	{
+		if (initializer.nonVoidInitializer &&
+			initializer.nonVoidInitializer.assignExpression &&
+			cast(UnaryExpression) initializer.nonVoidInitializer.assignExpression)
```
BBasile: One day we will be able to write 

```d
if (cast(UnaryExpression) initializer?.nonVoidInitializer?.assignExpression) {}
```
🙏

src/analysis/unmodified.d
```diff
@@ -211,6 +215,19 @@ private:
...
+		{
+			const UnaryExpression ue =
+				cast(UnaryExpression) initializer.nonVoidInitializer.assignExpression;
+			return ue.newExpression !is null;
+		}
+		else return false;
```
wilzbach: The else is superfluous

PR #539, opened 2017-12-03
Upload built binaries to GitHub

wilzbach:
For now Linux only. However, it can be used with [OSX too](https://github.com/jacob-carlborg/remarkify/blob/master/.travis.yml)

See also:
- https://docs.travis-ci.com/user/deployment/releases/
- https://docs.travis-ci.com/user/build-stages/deploy-github-releases

I will directly merge this as (1) we talked about this in the past and no one objected https://github.com/dlang-community/discussions/issues/9 and (2) it's an experiment anyways and (3) this just modifies the `.travis.yml` - no actual source code changes are happening.
I will also add a new tag afterwards to test this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,30 @@
 sudo: false
 language: d
 d:
-  - dmd-nightly
-  - dmd-beta
-  - dmd
-  - ldc-beta
-  - ldc
+- dmd-nightly
+- dmd-beta
+- dmd
+- ldc-beta
+- ldc
 os:
-  - linux
-  - osx
+- linux
+- osx
 env:
-  - BUILD=
-  - BUILD=dub
-
-script: ./.travis.sh
+- BUILD=
+- BUILD=dub
+script: "./.travis.sh"
+jobs:
+  include:
+  - stage: GitHub Release
+    d: ldc
+    os: linux
+    script: echo "Deploying to GitHub releases ..." && make ldcbuild
+    deploy:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,16 +1,30 @@
 sudo: false
 language: d
 d:
-  - dmd-nightly
-  - dmd-beta
-  - dmd
-  - ldc-beta
-  - ldc
+- dmd-nightly
```
wilzbach: In case someone wonders about this. I reformated the entire file with the official Travis CLI (this was needed to insert the `secret`).

wilzbach:
Alrighty - this is working fine now:

image

It will upload the release binary to GitHub for every new tag. If someone is interested, the same could be done for macOS. Building a binary for Windows might be trickier, but it doesn't seem that it will be too tricky:
- https://www.appveyor.com/docs/deployment/github/
- https://www.appveyor.com/docs/deployment/#deploy-on-tag-github-and-gitlab-only

WebFreak001:
shouldn't it rather create a .tar.xz instead of uploading a tarball and also rename it to linux x86_64?

wilzbach:
Why would we need to wrap it in an archive? It's a tiny binary anyways..
Yes, renaming based on the OS is on my list

WebFreak001:
well 5 MB isn't that tiny depending on your internet. Still better than electron apps I guess

PR #541, opened 2017-12-26
update dparse

BBasile:
The DUB tests wont work because the tag i required since a week now is still not pushed.


skl131313:
If you had remained part of the group, you could have done this as well. Expect it to be slower as there are now less people in the group and it is the holidays.

PR #546, opened 2018-01-09
Make splitting of lines for testing agnostic of a specific style.

skl131313:
So that `\n` can be used on Windows.

### DIFF
--- a/src/analysis/helpers.d
+++ b/src/analysis/helpers.d
@@ -61,7 +61,7 @@ void assertAnalyzerWarnings(string code, const StaticAnalysisConfig config,
 
 	// Run the code and get any warnings
 	MessageSet rawWarnings = analyze("test", m, config, moduleCache, tokens);
-	string[] codeLines = code.split(newline);
+	string[] codeLines = code.splitLines();
 
 	// Get the warnings ordered by line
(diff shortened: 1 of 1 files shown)


src/analysis/helpers.d
```diff
@@ -61,7 +61,7 @@ void assertAnalyzerWarnings(string code, const StaticAnalysisConfig config,
 
 	// Run the code and get any warnings
 	MessageSet rawWarnings = analyze("test", m, config, moduleCache, tokens);
-	string[] codeLines = code.split(newline);
+	string[] codeLines = code.splitLines();
```
Hackerpilot: I think the `import std.ascii : newline;` line can be deleted from this function as well.

Hackerpilot:
This is failing its tests on Windows 64-bit. I'm not really sure why though.

wilzbach:
```
Running .\bin\dscanner-test-library.exe 
Program exited with code -1073741819
```

Looks like a segfault. @skl131313 do you use Windows? Can you reproduce this?

skl131313:
Yah I get it, but I don't think it's caused by this pull request. It only happens half the time, the other half the program executes successfully. Seems to only be a problem with x64.

PR #547, opened 2018-01-18
Update LDC download url to use new format.

skl131313:
LDC stable has been released that uses the new URL format.


BBasile:
libdparse needs this too

BBasile:
Never mind, this will be done in my next PR very soon.

PR #549, opened 2018-01-25
Make visiting alias public

RazvanN7:
Making the visiting alias private makes it so that the visit(Module) overload is not visible outside the Outliner class. The code worked up until now due to a bug [1] in the compiler.

[1] https://github.com/dlang/dmd/pull/7766

### DIFF
--- a/src/outliner.d
+++ b/src/outliner.d
@@ -13,6 +13,8 @@ import std.conv;
 
 class Outliner : ASTVisitor
 {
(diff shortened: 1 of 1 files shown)


wilzbach:
The AppVeyor segfaults seem unrelated (appeared on previous builds too) & Travis is stalled due Travis's huge OSX bottleneck.

PR #555, opened 2018-02-01
fix #554 - ~master, infinite loop with -S and std.traits

### LINKED ISSUE #554: ~master, infinite loop with -S and std.traits
Maybe due to my last PR.

BBasile:


BBasile:
ping @skl131313, this is the same fix has the one you just merged in DCD.

PR #570, opened 2018-03-23
Add a temporary function that warns or patches the old INI section

BBasile:
After thinking more about the upcoming breaking change i think that finally the changelog and a forum announce could be not enough. This PR makes D-Scanner exit if the old INI section is found, with instruction about how to patch.

### DIFF
--- a/src/dscanner/main.d
+++ b/src/dscanner/main.d
@@ -68,6 +68,7 @@ else
 	bool printVersion;
 	bool explore;
 	string errorFormat;
+	bool patchConfig;
 
 	try
 	{
@@ -96,7 +97,8 @@ else
 				"muffinButton", &muffin,
 				"explore", &explore,
 				"skipTests", &skipTests,
-				"errorFormat|f", &errorFormat);
+				"errorFormat|f", &errorFormat,
+				"patchConfig", &patchConfig);
 		//dfmt on
 	}
 	catch (ConvException e)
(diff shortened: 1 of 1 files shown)


src/dscanner/main.d
```diff
@@ -463,3 +472,41 @@ string getConfigurationLocation()
 
 	return getDefaultConfigurationLocation();
 }
+
+
+/// Patch the INI file to v0.5.0 format.
+//TODO: remove this from v0.6.0
+bool hasWrongIniFileSection(string confiFilename, bool patch)
```
wilzbach: Typo: configFile

src/dscanner/main.d
```diff
@@ -463,3 +472,41 @@ string getConfigurationLocation()
...
+	import std.string : indexOf;
+	import std.array : replace;
+
+	bool result;
+
+	static immutable v1 = "analysis.config.StaticAnalysisConfig";
+	static immutable v2 = "dscanner.analysis.config.StaticAnalysisConfig";
+
+	char[] c = cast(char[]) readFile(confiFilename);
+	try if (const ptrdiff_t i = c.indexOf(v1))
```
wilzbach: BUG: v1 will _always_ be found. You need to check for v2 here.

wilzbach: -> https://github.com/dlang-community/D-Scanner/pull/572

PR #567, opened 2018-03-18
Allow customizing the style check messages' format

lenormf:
Hi,

This patch allows external tools to parse the diagnostics more easily. The code that compiles the format is the most stupid&simple I could think of, but maybe there's a more idiomatic way to do it.

Closes #377

PR #574, opened 2018-03-28
Don't force the documentation of the template parameter if it's a used type in the parameter list

wilzbach:
This relaxes the proper documentation checker to avoid enforcing the documentation of template parameters that are used as parameter types.
For example:

```
/++
An awesome description.

Params:
	r =  an input range.

Returns: Awesome values.
+/
string bar(R)(R r){}
```

The documentation of `R` isn't necessary here.
However, the following will trigger a warning about the missing documentation of `U`:

```
/++
An awesome description.

Params:
	r =  an input range.

Returns: Awesome values.
+/
string bar(U, R)(R r){}
```

CC @JackStouffer

### DIFF
--- a/src/dscanner/analysis/properly_documented_public_functions.d
+++ b/src/dscanner/analysis/properly_documented_public_functions.d
@@ -119,8 +119,8 @@ class ProperlyDocumentedPublicFunctions : BaseAnalyzer
 			return;
 
 		auto comment = setLastDdocParams(decl.name.line, decl.name.column, decl.comment);
-		checkDdocParams(decl.name.line, decl.name.column, decl.parameters);
-		checkDdocParams(decl.name.line, decl.name.column, decl.templateParameters);
+
+		checkDdocParams(decl.name.line, decl.name.column, decl.parameters,  decl.templateParameters);
 
 		enum voidType = tok!"void";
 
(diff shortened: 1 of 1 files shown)


src/dscanner/analysis/properly_documented_public_functions.d
```diff
@@ -187,36 +187,80 @@ private:
...
 			foreach (p; params.parameters)
 			{
+				string templateName;
+				if (auto t = p.type)
+				if (auto t2 = t.type2)
+				if (auto tip = t2.typeIdentifierPart)
+				if (auto iot = tip.identifierOrTemplateInstance)
+					templateName = iot.identifier.text;
+
+				auto idx = tlList.countUntil(templateName);
```
BBasile: `const` please

src/dscanner/analysis/properly_documented_public_functions.d
```diff
@@ -187,36 +187,80 @@ private:
 		return comment;
 	}
 
-	void checkDdocParams(size_t line, size_t column, const Parameters params)
+	void checkDdocParams(size_t line, size_t column, const Parameters params, const TemplateParameters templateParameters = null)
```
BBasile: new line for the `const Templ....` param

wilzbach:
Fixed. Also made a few other `auto`'s `const`

src/dscanner/analysis/properly_documented_public_functions.d
```diff
@@ -187,36 +187,81 @@ private:
...
-		import std.algorithm.searching : canFind;
+		import std.array : array;
+		import std.algorithm.searching : canFind, countUntil;
+		import std.algorithm.iteration : map;
+		import std.algorithm.mutation : remove;
+		import std.range : indexed, iota;
+
+		// convert templateParameters into a string[] for faster access
+		const(TemplateParameter)[] templateList;
+		if (const tp = templateParameters)
```
BBasile: `if (templateParameters && templateParameters.templateParameterList)`

src/dscanner/analysis/properly_documented_public_functions.d
```diff
@@ -187,36 +187,81 @@ private:
...
+		if (const tpl = tp.templateParameterList)
+			templateList = tpl.items;
+		string[] tlList;
+		tlList = templateList.map!(a => templateParamName(a)).array;
+
+		// make a copy of all parameters and remove the seen ones later during the loop
+		const(TemplateParameter)[] tpList;
+		if (const tp = templateParameters)
```
BBasile: `if (templateParameters && templateParameters.templateParameterList)`

src/dscanner/analysis/properly_documented_public_functions.d
```diff
@@ -187,36 +187,81 @@ private:
...
+		if (const tpl = tp.templateParameterList)
+			tpList = tpl.items;
+		size_t[] unseenTemplates = tpList.length.iota.array;
 
 		if (lastSeenFun.active && params !is null)
 			foreach (p; params.parameters)
 			{
+				string templateName;
+				if (const t = p.type)
```
BBasile: i'd use `&&` too here, instead of the 4 `if`

src/dscanner/analysis/properly_documented_public_functions.d
```diff
@@ -187,36 +187,81 @@ private:
...
+		if (const tpl = tp.templateParameterList)
+			templateList = tpl.items;
+		string[] tlList;
+		tlList = templateList.map!(a => templateParamName(a)).array;
+
+		// make a copy of all parameters and remove the seen ones later during the loop
+		const(TemplateParameter)[] tpList;
+		if (const tp = templateParameters)
+		if (const tpl = tp.templateParameterList)
+			tpList = tpl.items;
```
BBasile: `tpList` could be assigned at the same time as `templateList`. Also too bad it cant be copied (dup)

BBasile:
Something strange has happened while i started to review (review comments were posted but i haven't finished) but it's globally ok

src/dscanner/analysis/properly_documented_public_functions.d
BBasile: with Rebindable you could use `.dup`:  `Rebindable!(const TemplateParameter) tpList, templateList`;

wilzbach: `Rebindable` seems to be an alias for arrays and sadly doesn't work: 

```
src/dscanner/analysis/properly_documented_public_functions.d(208): Error: template object.dup cannot deduce function from argument types !()(const(TemplateParameter)[]), candidates are:
/usr/include/dlang/dmd/object.d(2086):        object.dup(T : V[K], K, V)(T aa)
/usr/include/dlang/dmd/object.d(2122):        object.dup(T : V[K], K, V)(T* aa)
/usr/include/dlang/dmd/object.d(4191):        object.dup(T)(T[] a) if (!is(const(T) : T))
/usr/include/dlang/dmd/object.d(4207):        object.dup(T)(const(T)[] a) if (is(const(T) : T))
```

src/dscanner/analysis/properly_documented_public_functions.d
wilzbach: Hmm imho the `if` are a lot clearer than:

```
if (p.type && p.type.type2 && p.type.type2.typeIdentifierPart &&
	p.type.type2.typeIdentifierPart.identifierOrTemplateInstance)
	templateName = p.type.type2.typeIdentifierPart.identifierOrTemplateInstance.identifier.text;
```

though I really miss `safeDeref` or `?` :/

src/dscanner/analysis/properly_documented_public_functions.d
wilzbach: it would then be 

```
if (templateParameters && templateParameters.templateParameterList)
    templateList = templateParameters.templateParameterList.items;
```

imho not much nicer.

src/dscanner/analysis/properly_documented_public_functions.d
BBasile: aka safe access operator, yes me too. These kind of OO structures would benefit of it...too bad the movement that started a few months (remember the thing with the Elvis operator ...) lead to nothing.

src/dscanner/analysis/properly_documented_public_functions.d
BBasile: strange because i tried earlier. I even asked on IRC.

BBasile: [Look at this small runnable](https://run.dlang.io/is/xGE1FL)

wilzbach: Interesting, but `remove` would still not as `remove` requires mutability of all elements (kind of stupid that you can't delete `const` elements, but well...)

src/dscanner/analysis/properly_documented_public_functions.d
BBasile: Maybe we should refactor D-Scanner [with this](https://github.com/BBasile/iz/blob/master/import/iz/sugar.d#L1560).

wilzbach: Yeah that would be quite helpful in many cases, but I'm not sure it's worth rewriting the existing code. Though maybe a good idea to add it to `utils` or something like that, s.t. the next time someone touches the code or implements a new check he can use it - or should we just add it then?

BBasile: yes in `utils`...making this and converting a few things PR will show quite directly if this is that useful.

wilzbach:
Looks like Travis is having problems with OSX jobs again:

image

I tend towards manually merging this now as 0.50 is coming up soonish - any last words?

BBasile:
Yes do it manually. They have updated the OSX infrastructure last week and since it's very hard to get picked from the queue...maybe they have bugs in their latest update.

Hackerpilot:
Looks like it's failing on Windows only, possibly due to D-Scanner still using an ancient containers library.

wilzbach:
Yeah looks like it happened on past PRs too:

https://github.com/dlang-community/D-Scanner/pull/567
https://github.com/dlang-community/D-Scanner/pull/570

(Not sure about the cause though.)

PR #577, opened 2018-03-30
add IZ safeAccess util and refactor several && chains with it

BBasile:


src/dscanner/analysis/static_if_else.d
```diff
@@ -47,17 +48,7 @@ class StaticIfElse : BaseAnalyzer
...
-			if (cc.falseStatement.statement.statementNoCaseNoDefault)
-			{
-				if (cc.falseStatement.statement.statementNoCaseNoDefault.ifStatement)
-				{
-					return cc.falseStatement.statement.statementNoCaseNoDefault.ifStatement;
-				}
-			}
-		}
-		return null;
+		return safeAccess(cc).falseStatement.statement.statementNoCaseNoDefault.ifStatement;
```
BBasile: This change is :1st_place_medal:

src/dscanner/analysis/mismatched_args.d
```diff
@@ -126,16 +127,16 @@ final class ArgVisitor : ASTVisitor
...
-			return;
-		if (unary.primaryExpression.identifierOrTemplateInstance is null)
-			return;
-		if (unary.primaryExpression.identifierOrTemplateInstance.identifier == tok!"")
-			return;
-		immutable t = unary.primaryExpression.identifierOrTemplateInstance.identifier;
-		lines ~= t.line;
-		columns ~= t.column;
-		args ~= internString(t.text);
+		if (const IdentifierOrTemplateInstance iot = safeAccess(unary)
```
wilzbach: Wouldn't `const iot` work too?

src/dscanner/analysis/static_if_else.d
wilzbach: Another bonus point for UFCS: `cc.safeAccess.falseStatement....` ;-)

src/dscanner/utils.d
```diff
@@ -0,0 +1,194 @@
...
+				static if (is(R == class))
+				{
+					return (m is null)
+						? SafeAccess!R(null)
+						: SafeAccess!R(__traits(getMember, m, member)(a));
+				}
+				else
+				{
+                    if (m)
+					    __traits(getMember, m, member)(a);
```
wilzbach: weird indentation style

src/dscanner/analysis/mismatched_args.d
BBasile: Yes. It can be removed unless DCD completion is needed.

BBasile: Wrong, the Type is actually required, otherwise the internal wrapper is returned.

BBasile: with `unwrap()` inference works

wilzbach: So why not use `unwrap` then?

BBasile: That's more o matter of style. I tend to use the type when i know it. Also better for DCD.

src/dscanner/utils.d
```diff
@@ -0,0 +1,194 @@
...
+		{
+			import std.traits : ReturnType, Parameters, isFunction;
+			static if (isFunction!T)
+			{
+				// otherwise there's a missing return statement.
+				alias R = ReturnType!T;
+				static if (!is(R == void) &&
+					!(is(R == class) && Parameters!T.length == 0))
+						pragma(msg, __FILE__ ~ "(" ~ __LINE__.stringof ~ "): error, " ~
+						"only `void function`s or `class` getters can be called without unwrap");
```
wilzbach: shouldn't this be an `static assert(0`, s.t. the compilation stops?
+ nit: use backticks with `unwrap`, s.t. DMD's colored highlighting works

wilzbach: Also why can't we return the result here?

BBasile: > shouldn't this be an static assert(0, s.t. the compilation stops?

the problem is that opDispatch gags the error messages so what happens exactly is that i let the error happens and use pragma to emit a pseudo error message.

PR #578, opened 2018-03-30
Use new config name override instead of breaking existing ini

BBasile:
We should be good for a release after that


BBasile:
Sorry @wilzbach i saw that you already patched the ini file in phobos but this is really what should have been done in first place.

wilzbach:
No biggie. I'm happy that this is fixed upstream :)

wilzbach:
BTW @BBasile I like your release announcement. As you did most of the work regarding 0.5, how about doing a quick announcement on the NG?

PR #580, opened 2018-03-30
Fix Make's shell to Bash

wilzbach:
So when I moved the new release bundling from the bash script into the Makefile, I "only" tested it locally and just assumed that it work on Travis too (as it currently does for dub).

It turns out that Make on Travis defaults to `/bin/sh` which points to `dash` on Ubuntu :/

See also: https://travis-ci.org/dlang-community/D-Scanner/jobs/360467457


BBasile:
Should i move the tag ? There was this stupid omission in the help message too...

PR #581, opened 2018-04-01
Adds a check for too much trusted scope, close #545

### LINKED ISSUE #545: Add a check that warns about mass-applied attributes
From https://github.com/dlang/phobos/pull/6010

tl;dr: this should be forbidden:

```d
@trusted:
```

BBasile:


src/dscanner/analysis/trust_too_much.d
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Checks that `@trusted` is only applied to a a single function
+ */
+class TrustTooMuchCheck : BaseAnalyzer
+{
+private:
+
+	static immutable MESSAGE = "Trusting a whole scope is a bad idea, " ~
+		"`@trusted` should only be attached to a single function";
+	static immutable string KEY = "dscanner.suspicious.catch_em_all";
```
WebFreak001: you reused an existing key here

BBasile:
@WebFreak001 ok fixed this key. BTW i don't think that KEYs are used at all ?

WebFreak001:
uh aren't they in the json response mode?

src/dscanner/analysis/config.d
```diff
@@ -195,6 +195,9 @@ struct StaticAnalysisConfig
 	@INI("Check indent of if constraints")
 	string if_constraints_indent = Check.disabled;
 
+    @INI("Check for @trusted applied to a bigger scope than a single function")
+    string trust_too_much = Check.enabled;
```
WebFreak001: I just noticed you used spaces instead of tabs here

BBasile: Yeah actually there was other spaces indents at the beginning, misleading my editor since it auto adapts to the current style.

src/dscanner/analysis/config.d
```diff
@@ -195,6 +194,9 @@ struct StaticAnalysisConfig
 	@INI("Check indent of if constraints")
 	string if_constraints_indent = Check.disabled;
 
+@INI("Check for @trusted applied to a bigger scope than a single function")
+string trust_too_much = Check.enabled;
```
wilzbach: Shouldn't there be indentation here?

src/dscanner/analysis/run.d
```diff
@@ -511,6 +512,10 @@ MessageSet analyze(string fileName, const Module m, const StaticAnalysisConfig a
 		checks ~= new IfConstraintsIndentCheck(fileName, tokens,
 		analysisConfig.if_constraints_indent == Check.skipTests && !ut);
 
+	if (moduleName.shouldRun!"trust_too_much"(analysisConfig))
+		checks ~= new TrustTooMuchCheck(fileName,
+		analysisConfig.trust_too_much == Check.skipTests && !ut);
```
wilzbach: BTW interested in reducing this boiler-plate with a `static foreach`?
I know that some classes use a different constructor, but we be trivial for most, e.g.

```
enum checks = [
   tuple("trust_too_much", TrustTooMuchCheck),
];
static foreach (check; checks)
...
```

src/dscanner/analysis/trust_too_much.d
```diff
@@ -0,0 +1,130 @@
...
+	@nogc @trusted { // [warn]: %s
+		void test();
+		void test();
+	}
+	}c.format(msg), sac);
+
+	assertAnalyzerWarnings(q{
+	@trusted template foo(){ // [warn]: %s
+	}
+	}c.format(msg), sac);
```
wilzbach: Maybe add one more test where `@trusted:` is used within an aggregate?

wilzbach:
Just for the record / those who are interested. Phobos isn't really koscher at the moment:

```
std/math.d(4924:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/math.d(5220:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/typecons.d(1957:6)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/typecons.d(5540:20)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/regex/package.d(482:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/regex/package.d(734:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/regex/package.d(1526:6)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/regex/internal/parser.d(544:6)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/regex/internal/ir.d(693:6)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/regex/internal/thompson.d(91:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/regex/internal/thompson.d(675:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/regex/internal/thompson.d(715:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/regex/internal/generator.d(13:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/regex/internal/kickstart.d(130:6)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/regex/internal/backtracking.d(16:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/stdio.d(3258:6)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/uni.d(3827:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/uni.d(4127:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/uni.d(8876:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
std/internal/cstring.d(185:2)[warn]: Trusting a whole scope is a bad idea, `@trusted` should only be attached to a single function
```

BBasile:
The std.math matches are not in 2.079. For the other maybe you'll have to edit the filters. The problem is that warnings in static analysis should rather have a psychological impact, i.e _look at there, there's possibly a problem_...rather than stopping the whole world. But again with the filtering system you've added last year problems listed here should be manageable.

BBasile:
I bet std.math is now affected because of the recent overloads added for the funcs that only accepted real...

wilzbach:
> The std.math matches are not in 2.079.

It was from the current master.

> But again with the filtering system you've added last year problems listed here should be manageable.

Yup, that's why it was just an FYI. The check is already very helpful and we can start with `-std.math,-std.uni,...` and then slowly weed out the whole scope `@trusted` abuses.

BBasile:
I'm checking the other now. I believe these are no false positive since the trust_too_much check is rather simple. I admit that i haven't tested phobos yesterday.

PR #582, opened 2018-04-02
Fix #268 - Use the User's profile directory under Windows

### LINKED ISSUE #268: Windows config file
Could you add the capability to have a config file in the userprofile directory?
For example...

``` d
string getConfigurationLocation()
{
    import std.process : environment;
    version (useXDG)
    {
        string configDir = environment.get("XDG_CONFIG_HOME", null);
        if (configDir is null)
        {
            configDir = environment.get("HOME", null);
            if (configDir is null)
                throw new Exception("Both $XDG_CONFIG_HOME and $HOME are unset");
            configDir = buildPath(configDir, ".config", "dscanner", CONFIG_FILE_NAME);
        }
        else
        {
            configDir = buildPath(configDir, "dscanner", CONFIG_FILE_NAME);
        }
        return configDir;
    }
    else version(Windows)
    {
        string configDir = environment.get("USERPROFILE", null);
        if (configDir is null)
            throw new Exception("$USERPROFILE is unset");
        configDir = buildPath(configDir, ".config", "dscanner", CONFIG_FILE_NAME);
        return configDir;
    }
}
```

wilzbach:
I don't use nor have Windows, but the code at https://github.com/dlang-community/D-Scanner/issues/268 looks reasonable.

And the variable seems to be always defined on Windows:

https://en.wikipedia.org/wiki/Environment_variable#Default_values

### DIFF
--- a/src/dscanner/main.d
+++ b/src/dscanner/main.d
@@ -407,24 +407,28 @@ version (OSX) version = useXDG;
  */
 string getDefaultConfigurationLocation()
 {
+	import std.process : environment;
+	import std.exception : enforce;
 	version (useXDG)
 	{
-		import std.process : environment;
-
 		string configDir = environment.get("XDG_CONFIG_HOME", null);
 		if (configDir is null)
 		{
 			configDir = environment.get("HOME", null);
-			if (configDir is null)
-				throw new Exception("Both $XDG_CONFIG_HOME and $HOME are unset");
+			enforce(configDir !is null, "Both $XDG_CONFIG_HOME and $HOME are unset");
 			configDir = buildPath(configDir, ".config", "dscanner", CONFIG_FILE_NAME);
 		}
 		else
 			configDir = buildPath(configDir, "dscanner", CONFIG_FILE_NAME);
 		return configDir;
 	}
-	else version (Windows)
-		return CONFIG_FILE_NAME;
+    else version(Windows)
+    {
+        string configDir = environment.get("APPDATA", null);
+        enforce(configDir !is null, "%APPDATA% is unset");
+        configDir = buildPath(configDir, "dscanner", CONFIG_FILE_NAME);
+        return configDir;
+    }
 }
 
 /**


src/dscanner/main.d
```diff
@@ -423,8 +422,14 @@ string getDefaultConfigurationLocation()
 			configDir = buildPath(configDir, "dscanner", CONFIG_FILE_NAME);
 		return configDir;
 	}
-	else version (Windows)
-		return CONFIG_FILE_NAME;
+    else version(Windows)
+    {
+        string configDir = environment.get("USERPROFILE", null);
+        if (configDir is null)
+            throw new Exception("$USERPROFILE is unset");
```
wilzbach: FWIW we could simply do:

```
environment.get("USERPROFILE");
```

if all we want is an exception if the variable doesn't exist.

skl131313: That the same thing, the default value passed i just a null, it only throws if there is an UTF error. According to the Doc at least.

https://github.com/dlang/phobos/blob/v2.079.0/std/process.d#L3375

Just a small nitpick, `$USERPROFILE` would be the bash way of declaring a variable. The windows way is either `$env:USERPROFILE` for power shell or `%USERPROFILE%` for cmd. I imagine most people will know what it means just by virtue of the name and caps.

BBasile: there's opIn now so finally you can do

`if (const string* = "USERPROFILE" in environment) {...}`

BBasile:
```diff
-        string configDir = environment.get("USERPROFILE", null);
+       string configDir = environment.get("APPDATA", null);
        if (configDir is null)
            throw new Exception("%APPDATA% is unset");
-        configDir = buildPath(configDir, ".config", "dscanner", CONFIG_FILE_NAME);
+        configDir = buildPath(configDir, "dscanner", CONFIG_FILE_NAME);
        return configDir;
```

src/dscanner/main.d
```diff
@@ -407,24 +407,28 @@ version (OSX) version = useXDG;
...
 		}
 		else
 			configDir = buildPath(configDir, "dscanner", CONFIG_FILE_NAME);
 		return configDir;
 	}
-	else version (Windows)
-		return CONFIG_FILE_NAME;
+    else version(Windows)
+    {
+        string configDir = environment.get("APPDATA", null);
```
wilzbach: > if all we want is an exception if the variable doesn't exist.

Sorry I was referring to environment["APPDATA"]:

https://github.com/dlang/phobos/blob/v2.079.0/std/process.d#L3335

I think a middle-ground is to use `enforce` as this is pretty explicit and saves one line.

PR #587, opened 2018-04-03
Don't use the dscanner prefix for the ModuleFilter in the config INI

wilzbach:
(we should probably tag 0.5.1 after this was merged)


BBasile:
Not wait for the tag, the milestone can be completed quickly https://github.com/dlang-community/D-Scanner/milestone/8

Hackerpilot:
I think now that we can actually specify section names we should pick ones that are shorter and more clear, but that's maybe something for 0.6.0.

PR #590, opened 2018-04-03
Use -flto=full for the LDC release build

wilzbach:
I wasn't sure whether we want to do this by default, so I set it only for the release build for now.
Though there isn't a huge time difference and we probably can add it to the `ldcbuild`.
Thoughts?

### DIFF
--- a/makefile
+++ b/makefile
@@ -27,6 +27,8 @@ VERSIONS =
 DEBUG_VERSIONS = -version=dparse_verbose
 DMD_FLAGS = -w -inline -release -O -J. -od${OBJ_DIR} -version=StdLoggerDisableWarning
 DMD_TEST_FLAGS = -w -g -J. -version=StdLoggerDisableWarning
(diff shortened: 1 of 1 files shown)


Hackerpilot:
I don't know enough about LDC's LTO to say that this is a good or bad idea.

wilzbach:
FYI: I just tried it with dub and it makes ~10% performance difference:

https://github.com/dlang/dub/pull/1433

Jon (from the ebay Search team) got similar results for his tsv-utils:

https://github.com/eBay/tsv-utils-dlang/blob/master/docs/dlang-meetup-14dec2017.pdf

My question was rather whether you want me to just add it to the LDC command (aka "default") or do it optionally only for the release build (as it's done now).

BBasile:
yes DUB use an archive for each lib, hence the diff. definitively ok.

BBasile:
Compile at once with make would be the same, this is what i meant.

Issue #591, opened 2018-04-03
DScanner dockerfile on Docker Hub


andre2007:
Do you already considered to provide a dockerfile on Docker Hub which allows to scan a project on a host system without a D / D-Scanner installation?

Use case is CI Systems like Jenkins(Pipelines) which clones git repositories and execute static code checks  on the projects.  In your pipeline you can execute docker (pseudo code example below) to scan your project. After the scan, the project dir will contain the D-Scanner result file. 

docker run -v ~/dProject:/data D-Scanner:latest -rm dscanner -s --report > report.json

An official docker hub repository from dlang-community would be quite convenient.

wilzbach:
Sure why not -> an initial attempt: https://github.com/dlang-community/D-Scanner/pull/592

andre2007:
This issue is solved.

PR #592, opened 2018-04-03
Auto-deploy to DockerHub

wilzbach:
An attempt at https://github.com/dlang-community/D-Scanner/issues/591

Busybox requires a fully-static build, but that's cool. The compressed size is 2MB.
Even Alpine has ~5MB.

Play with this:

https://hub.docker.com/r/dlangcommunity/dscanner

While it ideally would be:

```
docker run --rm -v $(pwd):/src dlangcommunity/dscanner --styleCheck .
```

This still results in a lot of "Could not resolve/locate location of module" errors, hence this mapping is required:

```
docker run --rm -v $(pwd):/src -v /usr/include/dlang/dmd:/d dlangcommunity/dscanner --styleCheck . -I/d
```

(dscanner is agnostic about what standard library you use)

I added dlangcommunitybot account and gave it permissions to push to dlangcommunity/dscanner + added the tokens to the Travis environment flags.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,24 @@ jobs:
           on:
             repo: dlang-community/D-Scanner
             tags: true
+    - stage: dockerhub-stable
+      if: tag IS present
+      d: ldc
+      os: linux
+      script:
+        - echo "Deploying to DockerHub..." && ./release.sh
+        - LATEST_TAG="$(git describe --abbrev=0 --tags)"
+        - docker build -t "dlangcommunity/dscanner:${LATEST_TAG} ."
+        - if [[ "$TRAVIS_BRANCH" == "master" && "$TRAVIS_PULL_REQUEST" == "false" ]] ; then docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD" ; fi
+        - if [[ "$TRAVIS_BRANCH" == "master" && "$TRAVIS_PULL_REQUEST" == "false" ]] ; then docker push "dlangcommunity/dscanner:${LATEST_TAG}" ; fi
+    - stage: dockerhub-latest
+      d: ldc
+      os: linux
+      script:
+        - echo "Deploying to DockerHub..." && ./release.sh
+        - docker build -t dlangcommunity/dscanner:latest .
+        - if [[ "$TRAVIS_BRANCH" == "master" && "$TRAVIS_PULL_REQUEST" == "false" ]] ; then docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD" ; fi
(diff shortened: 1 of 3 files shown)


andre2007:
Great you working on this. Thanks a lot. The deluxe solution would be if no dmd installation on host system is required. Maybe dmd could be added to the docker file.

wilzbach:
> The deluxe solution would be if no dmd installation on host system is required. Maybe dmd could be added to the docker file.

Not sure about this, because this would mean that 
- you may get different errors on your host as on the VM
- all the source location resolution would point to the docker instance
- this would require a more sophisticated release process for every "supported" D version, s.t. you can have Dscanner 0.5.0 + DMD 2.072.0

Can't you use the an existing Docker D image? (https://hub.docker.com/r/dlanguage/dmd/) Or extend upon this image and download the official source within it?

> Great you working on this. Thanks a lot. 

Well, I'm already done. I don't plan to invest more time as it's already working from my point of view. Once this is merged, new releases will auto-deployed to DockerHub without any maintenance required.

andre2007:
Thanks, as you proposed I will create a docker image which is based on this and will also download and install DMD.
In the CI infrastructure I have to use, I am not allowed to install any packages. Everything has to be provided by the docker image.

wilzbach:
> Thanks, as you proposed I will create a docker image which is based on this and will also download and install DMD.

Maybe the following is even better and easier then:

```docker
FROM dlanguage/dmd:2.079.0
ARG DSCANNER=0.5.1
ADD https://github.com/dlang-community/D-Scanner/releases/download/v${DSCANNER}/dscanner-v${DSCANNER}-linux-x86_64.tar.gz dscanner.tgz
RUN cat dscanner.tgz | tar -xzC /usr/bin && rm dscanner.tgz
```

Note: you can build an arbitrary Dscanner version without even touching the Dockerfile with this:

```sh
docker build --build-arg DSCANNER=0.5.0 -t dscanner:0.5.0 .
```

BBasile:
We just need this one to complete 0.5.2. Dont forget it, it's already 3 weeks old and i think you're the only one working on this kind of stuff in the community.

wilzbach:
Sorry - this one isn't really important to me, so feel free to release 0.5.2 without it.
I will eventually get to it tough.

wilzbach:
Finally managed to get to this and could figure out the cause of the failure.
For security reasons the environment variables aren't exposed in the pull request build, hence there's no an IF `TRAVIS_PULL_REQUEST`.
So this should finally work.
It's configured to always push to `latest`, but push a version tag on a git tag build.
Let's give this a try :)

wilzbach:
Looks like the auto-deployment works:

https://hub.docker.com/r/dlangcommunity/dscanner/tags/

PR #596, opened 2018-04-04
fix #195 - Unrecognized CLI options gives huge stacktrace

### LINKED ISSUE #195: Unrecognized CLI options gives huge stacktrace
`dscanner -a`: `Unrecognized option -a` followed by a long, long stacktrace. I thought the program crashed.

Looks like getopt just throws a regular exception rather than a specialized getopt exception.
I think adding a catch that looks for `Unrecognized option` would be nice.

BBasile:


BBasile:
This is trivial change actually. You can look at the stacktrace, it's of none interest.

Issue #575, opened 2018-03-29
Dscanner should allow opCmp with no toHash


JackStouffer:
Consider

```d
struct S
{
        int opCmp(S rhs)
        {
            assert(0);
        }
}

void main()
{
        assert(S.init == S.init);
}
```

The absence of opEquals means it's using the language default, which means it's safe for S to use the default toHash as well.

dscanner should only throw a warning when opEquals exists but toHash doesn't.

Hackerpilot:
I agree that this would be a good change.

PR #598, opened 2018-04-05
Improve the deployment script

wilzbach:
Testing https://github.com/dlang-community/discussions/issues/9#issuecomment-378725199

As I don't have OSX, I simply added `make release` to the Travis script for now. Let's see if this works out of the box.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,10 @@ script: "./.travis.sh"
 jobs:
   include:
     - stage: GitHub Release
-      if: tag IS present
-      d: ldc
+      #if: tag IS present
+      d: ldc-1.8.0
       os: linux
-      script: echo "Deploying to GitHub releases ..." && make release
+      script: echo "Deploying to GitHub releases ..." && ./release.sh
       deploy:
         provider: releases
         api_key: $GH_REPO_TOKEN
@@ -30,10 +30,10 @@ jobs:
           repo: dlang-community/D-Scanner
           tags: true
     - stage: GitHub Release
-      if: tag IS present
-      d: ldc
+      #if: tag IS present
+      d: ldc-1.8.0
       os: osx
-      script: echo "Deploying to GitHub releases ..." && make release
+      script: echo "Deploying to GitHub releases ..." && ./release.sh
       deploy:
         provider: releases
         api_key: $GH_REPO_TOKEN
--- a/makefile
+++ b/makefile
@@ -76,26 +76,5 @@ report: all
 	dscanner --report src > src/dscanner-report.json
 	sonar-runner
 
-.ONESHELL:
 release:
-	@set -eux -o pipefail
-	VERSION=$$(git describe --abbrev=0 --tags)
-	ARCH="$${ARCH:-64}"
-	unameOut="$$(uname -s)"
-	case "$$unameOut" in
-	    Linux*) OS=linux; ;;
-	    Darwin*) OS=osx; ;;
-	    *) echo "Unknown OS: $$unameOut"; exit 1
-	esac
-
-	case "$$ARCH" in
-	    64) ARCH_SUFFIX="x86_64";;
-	    32) ARCH_SUFFIX="x86";;
-	    *) echo "Unknown ARCH: $$ARCH"; exit 1
-	esac
-
-	archiveName="dscanner-$$VERSION-$$OS-$$ARCH_SUFFIX.tar.gz"
-
-	echo "Building $$archiveName"
-	${MAKE} ldcbuild LDC_FLAGS="-flto=full"
-	tar cvfz "bin/$$archiveName" -C bin dscanner
+	./release.sh
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+set -eux -o pipefail
+VERSION=$(git describe --abbrev=0 --tags)
+ARCH="${ARCH:-64}"
+LDC_FLAGS=()
+unameOut="$(uname -s)"
+case "$unameOut" in
+    Linux*) OS=linux; LDC_FLAGS=("-flto=full" "-linker=gold" "-static") ;;
+    Darwin*) OS=osx; LDC_FLAGS+=("-L-macosx_version_min" "-L10.7" "-L-lcrt1.o"); ;;
+    *) echo "Unknown OS: $unameOut"; exit 1
+esac
+
+case "$ARCH" in
+    64) ARCH_SUFFIX="x86_64";;
+    32) ARCH_SUFFIX="x86";;
+    *) echo "Unknown ARCH: $ARCH"; exit 1
+esac
+
+archiveName="dscanner-$VERSION-$OS-$ARCH_SUFFIX.tar.gz"
+
+echo "Building $archiveName"
+${MAKE:-make} ldcbuild LDC_FLAGS="${LDC_FLAGS[*]}"
+tar cvfz "bin/$archiveName" -C bin dscanner


release.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -eux -o pipefail
+VERSION=$(git describe --abbrev=0 --tags)
+ARCH="${ARCH:-64}"
+LDC_FLAGS=("-flto=full")
+unameOut="$(uname -s)"
+case "$unameOut" in
+    Linux*) OS=linux; ;;
+    Darwin*) OS=osx; LDC_FLAGS+=("-macosx_version_min 10.7" "-lcrt1.o"); ;;
```
wilzbach:
```
ldc2: Unknown command line argument '-macosx_version_min'.  Try: 'ldc2 -help'
ldc2: Did you mean '-dwarf-version'?
ldc2: Unknown command line argument '-lcrt1.o'.  Try: 'ldc2 -help'
```

:/

jacob-carlborg: These are linker flags. You can check the DStep Dub file.

makefile
```diff
@@ -76,26 +76,5 @@ report: all
 	dscanner --report src > src/dscanner-report.json
 	sonar-runner
 
-.ONESHELL:
```
wilzbach: `ONESHELL` is only available in [Make 3.82](https://lists.gnu.org/archive/html/info-gnu/2010-07/msg00023.html).
Travis uses 3.81 and there's no easy way to upgrade.

.travis.yml
```diff
@@ -16,10 +16,10 @@ script: "./.travis.sh"
 jobs:
   include:
     - stage: GitHub Release
-      if: tag IS present
-      d: ldc
+      #if: tag IS present
```
wilzbach: Apparently building the binaries on Travis is vastly more complicated than thought, so let's better constantly build them to avoid bad surprises on releases. Imho the ~1.5min are worth the mind peace.

.travis.yml
```diff
@@ -16,10 +16,10 @@ script: "./.travis.sh"
...
+      d: ldc-1.8.0
```
wilzbach: Locking the release compiler. Just in case there are any regressions coming up (I experienced some with the dlang-tour and it was a hell to figure out that the sudden failures were due to an automatic compiler upgrade)

wilzbach:
@BBasile FYI: this is ready now. Would be nice if we can merge this _before_ tagging 0.5.1 - thanks!

release.sh
```diff
@@ -0,0 +1,23 @@
...
+    *) echo "Unknown OS: $unameOut"; exit 1
+esac
+
+case "$ARCH" in
+    64) ARCH_SUFFIX="x86_64";;
+    32) ARCH_SUFFIX="x86";;
+    *) echo "Unknown ARCH: $ARCH"; exit 1
+esac
+
+archiveName="dscanner-$VERSION-$OS-$ARCH_SUFFIX.tar.gz"
```
jacob-carlborg: Nitpick, binaries/packages for macOS usually don't contain the architecture. If multiple architectures are provided it's expected to be provided in a single universal binary.

release.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -eux -o pipefail
+VERSION=$(git describe --abbrev=0 --tags)
+ARCH="${ARCH:-64}"
+LDC_FLAGS=()
+unameOut="$(uname -s)"
+case "$unameOut" in
+    Linux*) OS=linux; LDC_FLAGS=("-flto=full" "-linker=gold" "-static") ;;
+    Darwin*) OS=osx; LDC_FLAGS+=("-L-macosx_version_min" "-L10.7" "-L-lcrt1.o"); ;;
```
jacob-carlborg: Nitpick, I would prefer "macOS" over `osx`, since this is what the OS is called now.

BBasile:
@wilzbach there's still #575 that's planned for 0.5.1. It's simple, will do it today, the tag will follow then.

release.sh
wilzbach: Hmm, we do the same for dub too (https://github.com/dlang/dub/releases, http://code.dlang.org/download) as it makes automated downloads easier.
I just saw that LDC has the same pattern: https://github.com/ldc-developers/ldc/releases

release.sh
wilzbach: The naming is consistent with Dub's and LDC's releases:

https://github.com/dlang/dub/releases
https://github.com/ldc-developers/ldc/releases

wilzbach:
@BBasile I will merge this then, s.t. we can test it for 0.5.1
@jacob-carlborg If you feel strongly about the package name, we can always change it for 0.5.2 ;-)

release.sh
BBasile: Yes, please @wilzbach. He right about the terminology. osx is a serie of  Mac OS versions.

BBasile: But that's true that even for pre-defined D version identifiers osx is used....

wilzbach: Well, to be honest I don't feel strongly about it as long as we finally get the automated release builds working. We have been pushing this off for far too long ;-)

wilzbach:
Looks like this finally worked :)

image

I tested the Linux binary and it works fine - even on a 2.6.X kernel 

So now we just need to figure out how to painlessly compile the Windows binaries.
Did someone try the Wine-built binary of DScanner that I posted [here](https://github.com/dlang-community/discussions/issues/9#issuecomment-378440299)? If that works, it might be a feasible approach.
Otherwise I LDC supports [cross-compiling since a while](https://github.com/ldc-developers/ldc/pull/2142#issuecomment-304472412) though it might require some work to get that working flawlessly in Docker (I saw on Slack that @jacob-carlborg gave this a shot recently) + there's the licensing problem.

jacob-carlborg:
> Otherwise I LDC supports cross-compiling since a while though it might require some work to get that working flawlessly in Docker (I saw on Slack that jacob-carlborg gave this a shot recently) + there's the licensing problem.

You don't need to run it in Docker. It simple enough to setup everything that is needed before building. This Docker image, compared with the one for macOS, did not require the multiarch/crossbuild base Docker image since LDC contains the cross-linker.

jacob-carlborg:
> If you feel strongly about the package name

I don't.

release.sh
jacob-carlborg: > The naming is consistent with Dub's and LDC's releases:
But that's true that even for pre-defined D version identifiers osx is used....

That doesn't mean it's correct :smiley:. The predefined version identifier was arbitrary chosen by Walter even though `darwin` already existed in other compilers. Back then the official name was "Mac OS X", so naming it "OSX" was neither correct.

jacob-carlborg:
Adding `-L-dead_strip` on macOS will reduce the binary size to almost a third. This corresponds to the `--gc-sections` flag that LDC uses by default on Linux. Not sure why the macOS flag is not enabled by default.

Adding `-flto=full` and running `strip` on the binary will get the size down to 3MB from the original 14MB.

wilzbach:
> You don't need to run it in Docker. 

I know, but I was referring to downloading the VS libraries from Dropbox and copying the libs over.
That seems a bit fragile and like something I don't want to do for all X repositories as it seems like it might require small tweaks in the future.
I'm already not very well pleased about the duplication of the build script between D-Scanner, DCD and dfmt and thinking about moving the build script to tools ;-)

jacob-carlborg:
Right. Perhaps a pre-packaged cross-compiler? Download and unpack. Event better if it can be upstreamed.

wilzbach:
> @wilzbach If you don't want to learn PowerShell, you can use AppVeyor and install msys2 which bundles bash. I did some experiments and it's pretty straight forward [1]:

Thanks a lot for your pointers, but we already went through this and while annoying PowerShell isn't too hard. Even this repo has a `appveyor.yml` that builds and uploads the binary:

https://github.com/dlang-community/D-Scanner/blob/master/appveyor.yml

But AppVeyor was disabled, because I think (?) there have been too many random failures.
For now, we went with using Travis and Wine to build the Windows binaries:

https://github.com/dlang-community/D-Scanner/pull/603

PR #604, opened 2018-04-06
Use git --describe tags as version string

wilzbach:
I just noticed that DScanner's version number doesn't get bumped and still pointed to 0.4.0 (rather confusing.)

As manually bumped the version is error-prone (i.e. it gets forgotten), I suggest to use `git describe --tags` wherever we can.

```
> ./bin/dscanner --version
v0.5.1-2-g3612c01
```

```
> git stash && git checkout v0.5.1 && git stash apply
> make
> ./bin/dscanner --version
v0.5.1-2-g3612c01
```

### DIFF
--- a/makefile
+++ b/makefile
@@ -36,7 +36,7 @@ ldc: ldcbuild
 gdc: gdcbuild
 
 githash:
-	git log -1 --format="%H" > githash.txt
+	git describe --tags > githash.txt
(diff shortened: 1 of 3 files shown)


wilzbach:
@BBasile @Hackerpilot are you okay with this?

PR #605, opened 2018-04-06
Generate a valid version for the DUB build

wilzbach:
A hack at exposing the package's version for the DUB build.
It will default to `git describe --tags` (within the packages's directory) and if that fails it will detect its version based on the DUB package path.
If that fails too, it will fallback to `0.0.0`.

See also:

https://github.com/dlang-community/D-Scanner/pull/604 (which is about the outdated version)

### DIFF
--- a/src/dscanner/dscanner_version.d
+++ b/src/dscanner/dscanner_version.d
@@ -10,10 +10,11 @@ module dscanner.dscanner_version;
  */
 enum DSCANNER_VERSION = "v0.4.0";
 
-version (Windows)
+version (built_with_dub)
 {
+	enum GIT_HASH = import("dubhash.txt");
 }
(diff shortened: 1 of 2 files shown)


src/dscanner/dscanner_version.d
```diff
@@ -10,10 +10,11 @@ module dscanner.dscanner_version;
  */
 enum DSCANNER_VERSION = "v0.4.0";
 
-version (Windows)
+version (built_with_dub)
 {
+	enum GIT_HASH = import("dubhash.txt");
```
wilzbach: FYI: Left it like this for now - it will conflict with https://github.com/dlang-community/D-Scanner/pull/605 anyhow.

PR #611, opened 2018-04-12
Change trust_too_much message which was slightly ambiguous

BBasile:
It could be interpreted as "use it once in the whole scope"


BBasile:
By the way, did you activate it for phobos ? D. Nadlinger just found something that would have been detected (https://issues.dlang.org/show_bug.cgi?id=18755)

wilzbach:
Yes, see:

https://github.com/dlang/phobos/pull/6446
https://github.com/dlang/phobos/pull/6447

BBasile:
okay nice, but Rebindable "get" should not be trusted i think. Anyway off topic

PR #621, opened 2018-04-28
Fix pre-generate command on Windows

LaurentTreguier:
Using single quotes instead of double quotes seem to pose a problem on my Windows system for some reason... I end up with this when trying to compile it using dub:
```
object.Exception@rdmd.d(191): Cannot have both --eval and a program file ('dir=environment.get(`DUB_PACKAGE_DIR`);').
```

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -24,6 +24,6 @@
     "bin"
   ],
   "preGenerateCommands" : [
-    "rdmd --eval='auto dir=environment.get(`DUB_PACKAGE_DIR`); dir.buildPath(`bin`).mkdirRecurse; auto gitVer = (`git -C `~dir~` describe --tags`).executeShell; (gitVer.status == 0 ? gitVer.output : dir.dirName.baseName.findSplitAfter(environment.get(`DUB_ROOT_PACKAGE`)~`-`)[1]).ifThrown(`0.0.0`).toFile(dir.buildPath(`bin`, `dubhash.txt`));'"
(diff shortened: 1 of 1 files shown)


LaurentTreguier:
I didn't think that backquotes would be interpretated inside double quotes, so Posix builds are failing now.

wilzbach:
> Using single quotes instead of double quotes seem to pose a problem on my Windows system for some reason

That's really weird indeed as the entire string should be passed to `rdmd --eval`. Do you by chance have any idea what's going wrong (I'm not a Windows user)?

LaurentTreguier:
> Nope, it shouldn't. That's why there are two fallbacks.

Indeed, I should have actually read the line entirely before saying anything.

> Do you by chance have any idea what's going wrong ?

I think it's simply because on Windows' `cmd`, single quotes do not delimit strings, which means that 1) they are part of the string given to `--eval`, and 2) they don't even encapsulate spaces. Only double quotes seem to be interprated as string delimiters.

Cody-Duncan:
Tried this fix out locally. Build succeeds quite nicely. 

For anyone else on windows that's hitting this issue as a result of a dependency on D-Scanner. Here's a batch file I wrote to pull down this fix for your local dub package:

```
set current_dir=%cd%

REM go to the dub package directory
cd %AppData%\dub\packages\dscanner-0.5.2\dscanner\

REM turn it into a repository clone
git init
git remote add origin https://github.com/dlang-community/D-Scanner.git
git fetch
git reset --hard origin/master
git checkout -b origin/master

REM get the fix
git pull origin pull/621/head:fix_issue_621

REM return home
cd %current_dir%
```

PR #624, opened 2018-05-04
Update dependencies - possible fix for D Bugzilla 18667

BBasile:
I've encountered an issue similar to https://issues.dlang.org/show_bug.cgi?id=18667 in the last fix involving D-Symbol and DCD. The problem only happened with DMD 2.080 and probably with ~master too.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -13,9 +13,9 @@
   ],
   "dependencies" : {
     "libdparse" : "~>0.8.1",
-    "dsymbol" : "~>0.3.1",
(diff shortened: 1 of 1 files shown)


BBasile:
This is a workaround of course, real fix needs to be done in phobos.

BBasile:
FYI information, [this was the back trace](https://github.com/dlang-community/DCD/pull/476#issuecomment-385760771) and it looks __VERY__ similar.

PR #638, opened 2018-05-22
update version number

BBasile:


wilzbach:
Is this even necessary here?
I thought we recently fixed the dub.json and Makefile to auto-generate this?

BBasile:
Right but under windows the build.bat still uses the hardcoded version.

PR #639, opened 2018-05-24
fix #635 - Check for documented `throw` could also detect calls to `enforce`

### LINKED ISSUE #635: Check for documented `throw` could also detect calls to `enforce`

BBasile:


src/dscanner/analysis/properly_documented_public_functions.d
```diff
@@ -51,6 +51,49 @@ final class ProperlyDocumentedPublicFunctions : BaseAnalyzer
...
+            tt.type2 = new Type2;
+            tt.type2.typeIdentifierPart = new TypeIdentifierPart;
+            tt.type2.typeIdentifierPart.identifierOrTemplateInstance =
+                new IdentifierOrTemplateInstance;
+            tt.type2.typeIdentifierPart.identifierOrTemplateInstance.identifier =
+                Token(tok!"identifier", "Exception", 0, 0, 0);
+            thrown ~= tt;
+        }
+        // enforce!Type(condition);
+        else if (const TemplateSingleArgument tsa = safeAccess(iot).templateInstance
```
Hackerpilot: This branch doesn't seem to check for "enforce".

BBasile: Yes fixed and thanks for finding this.

BBasile:
I should probably check if std.exception is imported but this adds directly more complexity since the scope is needed.

PR #646, opened 2018-06-02
fix #645 - False positive, Else branch identical to Then branch with incomplete IfStatement

### LINKED ISSUE #645: False positive, Else branch identical to Then branch with incomplete IfStatement
```d
void main()
{
    if (stuff s = call())
}
```

BBasile:
When writing code both the _then_ and the _else_ can be null


BBasile:
This PR needs the fix for #644.

PR #650, opened 2018-06-04
Fix #648 - Use the current git version when building the Windows binaries

### LINKED ISSUE #648: Wrong version: --version outputs v0.5.5 instead of v0.5.6
D-Scanner windows release v0.5.6 outputs version v0.5.5 instead of v0.5.6

wilzbach:


build.bat
```diff
@@ -4,7 +4,19 @@ setlocal enabledelayedexpansion
...
+:: the existing githash file if existent
+git describe --tags > githash_.txt
+for /f %%i in ("githash_.txt") do set githashsize=%%~zi
+if %githashsize% == 0 (
+	if not exist "githash.txt" (
+		echo "0.0.0" > githash.txt
+	)
+) else (
+	move /y githash_.txt githash.txt
+)
```
wilzbach: We can't assume `git` to be installed, so hence this ugly workaround with fallback and not overwriting githash.txt if it already exists.
This works for TravisCi and should allow building DScanner on any existing Windows setup

build.bat
```diff
@@ -4,7 +4,19 @@ setlocal enabledelayedexpansion
...
 if "%DC%"=="ldc2" set DC="ldmd2"
 
-set DFLAGS=-O -release -inline -version=StdLoggerDisableWarning
+:: git might not be installed, so we provide 0.0.0 as a fallback or use
+:: the existing githash file if existent
+git describe --tags > githash_.txt
+for /f %%i in ("githash_.txt") do set githashsize=%%~zi
+if %githashsize% == 0 (
+	if not exist "githash.txt" (
+		echo "0.0.0" > githash.txt
```
LaurentTreguier: Just tested under wine, for some reason the printed version after compiling when git is not installed is `"0.0.0"` (with the quotes, instead of just `0.0.0`)

src/dscanner/dscanner_version.d
```diff
@@ -8,16 +8,11 @@ module dscanner.dscanner_version;
 /**
  * Human-readable version number
  */
-enum DEFAUULT_DSCANNER_VERSION = "v0.5.5";
```
Hackerpilot: How did none of us notice that this was spelled incorrectly?

PR #655, opened 2018-06-08
Add check for pointless arithmetic

kubo39:


BBasile:
You can really put everything now.

- binary op:
  - x^^0
  - x*1
- same for opOpAssign:
  - /=1
  - %= 1
  - *= 1

src/dscanner/analysis/pointless_arithmetic_check.d
```diff
@@ -0,0 +1,93 @@
...
+	{
+		super(fileName, null, skipTest);
+	}
+
+	override void visit(const MulExpression mulExp)
+	{
+		UnaryExpression left = cast(UnaryExpression) mulExp.left;
+		UnaryExpression right = cast(UnaryExpression) mulExp.right;
+
+		if (mulExp.operator == tok!"/" &&
```
BBasile: Please protect with `if (left && right){ ... }`

kubo39: oh, I missed. Thank you.

src/dscanner/analysis/pointless_arithmetic_check.d
```diff
@@ -0,0 +1,165 @@
...
+					 right.primaryExpression.primary.text == "1")
+			{
+				addErrorMessage(mulExp.line, mulExp.column, KEY,
+								"Any number modulo 1 will be 0.");
+			}
+		}
+		mulExp.accept(this);
+	}
+
+	static bool isIntegerLiteral(IdType type) nothrow pure
```
BBasile: Please move this in the `private:` section. 
Also this could go in dparse in a further PR but for now it's ok as a local helper.

src/dscanner/analysis/pointless_arithmetic_check.d
```diff
@@ -0,0 +1,165 @@
+//          Copyright Brian Schott (Hackerpilot) 2016.
```
Hackerpilot: I'm not sure putting my name on this makes sense. I didn't write this file.

kubo39:
This is meant to eliminate redundancy, not for performance. Nobody will write such code deliberately.

BBasile:
>  Nobody will write such code deliberately

Actually i'm not sure anymore. For example yesterday i've discovered a trick that does pointless stuff to promote a 1 or 2 bytes int:

```d
ubyte b1;
auto i1 = b1 + 0;
static assert(is(typeof(i1) == int));
```

BBasile:
I think that this won't gonna make it finally . Sorry if i encouraged you to loose time on this new check but since you seem to be able there are many issues and even proposal for new checks here: https://github.com/dlang-community/D-Scanner/issues. Also if you want to make a new check it will be safer to open an issue before, so that it can be discussed.

PR #688, opened 2018-08-22
Disable useless init chek by default

BBasile:
This one is more for consistency because people sometimes mix the two styles. More advanced D programmers may enable this check because they know the language better.


wilzbach:
I was actually thinking about the default value of DScanner:

https://github.com/dlang-community/D-Scanner/blob/44e0092b88d426708096b718d0c5680eaa7b09ba/src/dscanner/analysis/config.d#L179-L180

BBasile:
Damn, this is what i wished to do too.

PR #703, opened 2018-09-16
fix #696 - Unused variable fake positive with UFCS

### LINKED ISSUE #696: Unused variable fake positive with UFCS
```d
void main()
{
    import std.stdio : writeln;

    const int testValue; // warning for never used variable
    testValue.writeln; // if you write testValue.writeln() instead, you get no warning.
}
```

from https://github.com/BBasile/Coedit/issues/358

BBasile:
Apparently it was a bug not only specific to the UFCS.
Now every use of the member access operator is handled.


BBasile:
do not merge will update dsymbol.

BBasile:
dsymbol was for another PR that's already merged

PR #720, opened 2018-11-26
Update dparse to get latest feature and fix for phobos linting

BBasile:
This fixes problems with __vector, empty asm statement and __traits used as type (i'm a bit speculative with this one ... )


BBasile:
green, please review and accept if applicable.

PR #728, opened 2018-11-29
fix - adaptation to correct DIP 1009 impl. was incomplete

BBasile:
leading to phobos linting problems

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -13,5 +13,5 @@ else
     make lint
     git clone https://www.github.com/dlang/phobos.git --depth=1
     # just check that it doesn't crash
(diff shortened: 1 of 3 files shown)


BBasile:
D-Scanner CI test phobos too now so we should see if it's good without adding more tests

wilzbach:
> D-Scanner CI test phobos too now so we should see if it's good without adding more tests

I would at least add one basic test as its easy and will make debugging easier if the check needs to be improved / reworked at some point.

BBasile:
Perfect, look at first DMD log on Travis, no more warning.

PR #737, opened 2018-12-29
update dependencies to allow building with `-dip25`

BBasile:


LaurentTreguier:
If the `libdparse` git submodule is updated, maybe it should be also be updated in `dub.json` to keep them in sync

BBasile:
Yes i think it was updated by error. The new version doesn't fix anything anyway.

PR #623, opened 2018-04-29
Build win64 binaries

wilzbach:
I already uploaded them for the last release and tested them in a VirtualBox.
It's really just as easy as adding `-m64`.

PR #743, opened 2019-01-04
Use LDC for cross-compilation Win{32,64} binaries

wilzbach:
Follow-up to https://github.com/dlang-community/D-Scanner/pull/623

Based on @BBasile's comment: https://github.com/dlang-community/D-Scanner/pull/623#issue-184847942

### DIFF
--- a/makefile
+++ b/makefile
@@ -26,6 +26,9 @@ INCLUDE_PATHS = \
 VERSIONS =
 DEBUG_VERSIONS = -version=dparse_verbose
 DMD_FLAGS = -w -release -O -Jbin -od${OBJ_DIR} -version=StdLoggerDisableWarning
+override DMD_FLAGS += $(DFLAGS)
+override LDC_FLAGS += $(DFLAGS)
+override GDC_FLAGS += $(DFLAGS)
 DMD_TEST_FLAGS = -w -g -Jbin -version=StdLoggerDisableWarning
 override LDC_FLAGS += -O5 -release -oq
 override GDC_FLAGS += -O3 -frelease
--- a/release-windows.sh
+++ b/release-windows.sh
@@ -3,24 +3,52 @@
 set -eux -o pipefail
 VERSION=$(git describe --abbrev=0 --tags)
 OS=windows
-ARCH_SUFFIX="x86"
+if [ "${ARCH:-32}" == "64" ] ; then
+	ARCH_SUFFIX="x86_64"
+	export DFLAGS=-mtriple=x86_64-windows-msvc
+else
+	ARCH_SUFFIX="x86"
+	export DFLAGS=-mtriple=x86_32-windows-msvc
+fi
+
+# for the install.sh script only
+LDC_PATH="$(dirname $(dirname $(which ldc2)))"
 
 # Allow the script to be run from anywhere
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $DIR
 
-# Step 1: download the DMD binaries
-if [ ! -d dmd2 ] ; then
-	wget http://downloads.dlang.org/releases/2.x/2.079.0/dmd.2.079.0.windows.7z
-	7z x dmd.2.079.0.windows.7z > /dev/null
+# Step 1: download the LDC windows binaries
+if [ ! -d ldc2-1.13.0-windows-x64 ] ; then
+	wget https://github.com/ldc-developers/ldc/releases/download/v1.13.0/ldc2-1.13.0-windows-x64.7z
+	7z x ldc2-1.13.0-windows-x64.7z > /dev/null
 fi
 
-# Step 2: Run DMD via wineconsole
+# Step 2: Add LDC windows binaries to LDC Linux
+if [ ! -d "${LDC_PATH}/lib-win64" ] ; then
+	cp -r ldc2-1.13.0-windows-x64/lib "${LDC_PATH}/lib-win64"
+
+	cat >> "$LDC_PATH"/etc/ldc2.conf <<EOF
+"x86_64-.*-windows-msvc":
+{
+	switches = [
+		"-defaultlib=phobos2-ldc,druntime-ldc",
+		"-link-defaultlib-shared=false",
+	];
+	lib-dirs = [
+		"%%ldcbinarypath%%/../lib-win64",
+	];
+};
+EOF
+
+fi
+
+# Step 3: Run LDC with cross-compilation
 archiveName="dscanner-$VERSION-$OS-$ARCH_SUFFIX.zip"
 echo "Building $archiveName"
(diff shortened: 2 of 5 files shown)


BBasile:
As said in the other PR, there's no need to use the DFLAG as i did when experimenting. Add this to the makefile

```make
ldcxwin64:
	${LDC} ${LDC_FLAGS} -of=bin/dscanner ${VERSIONS} ${INCLUDE_PATHS} ${SRC} -Jbin -mtriple=x86_64-windows-msvc
```

And then oviously just `make ldcxwin64` is enough to cross compile.

wilzbach:
This PR will now use LDC for cross-compilation for both x86 and x86_64 and produce these binaries:

[dscanner-v0.5.7-windows-x86_64.zip](https://github.com/dlang-community/D-Scanner/files/2728013/dscanner-v0.5.7-windows-x86_64.zip)
[dscanner-v0.5.7-windows-x86.zip](https://github.com/dlang-community/D-Scanner/files/2728014/dscanner-v0.5.7-windows-x86.zip)

It would be great to get feedback from a Windows user on these binaries.

wilzbach:
> As said in the other PR, there's no need to use the DFLAG as i did when experimenting. Add this to the makefile
> 
> ```makefile
> ldcxwin64:
> 	${LDC} ${LDC_FLAGS} -of=bin/dscanner ${VERSIONS} ${INCLUDE_PATHS} ${SRC} -Jbin -mtriple=x86_64-windows-msvc
> ```
> 
> And then oviously just `make ldcxwin64` is enough to cross compile.

Let's do beautification at the end (i.e. once Travis is green) and in this case actually using `DFLAGS` isn't too bad because then I can use LDC for cross-compilation to 32-bit too.

wilzbach:
Yup, this is also working on Travis :tada: 

image

BBasile:
I'll go on win to test them...

wilzbach:
Thanks!
I guess now the only remaining bit would be to ensure that all LDC release flags are used for the binaries (the current Makefiles setup doesn't use any).

wilzbach:
So apparently there was some old binaries in my `bin` directory and it could be possible that the script packaged those for the zip archive.
Would you mind verifying these binaries again:

[dscanner-v0.6.0-windows-x86_64.zip](https://github.com/dlang-community/D-Scanner/files/2728258/dscanner-v0.6.0-windows-x86_64.zip)
[dscanner-v0.6.0-windows-x86.zip](https://github.com/dlang-community/D-Scanner/files/2728259/dscanner-v0.6.0-windows-x86.zip)

(this now also includes LDC with all optimization levels except for full LTO as this is probably a lot harder to get to work with cross-compilation).

wilzbach:
> I let you merge when you're happy with the PR.

Yup, I am happy now. We just need to make once sure that all these binaries are properly built.

> Then DCD and DFMT will need this change too.

I split up the LDC setup in a separate file, s.t. it's easier to copy the file over (though on the long run we might want to add stuff like this to a shared `tools` repo ...)
Anyhow:

https://github.com/dlang-community/DCD/pull/568
https://github.com/dlang-community/dfmt/pull/420

wilzbach:
Hmm so I was testing this in one of the [free VirtualBox for Windows IE](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/), but it complains about the missing `vcruntime140.dll` here:

image

And wine just aborts with "wine: Bad EXE format for ". So I guess we must have missed sth. here.

BBasile:
> So apparently there was some old binaries

Yes i should have seen that 0.5.7 was not the latest version

> We just need to make once sure that all these binaries are properly built.

They work here. The Only issue i see is that there a missing switch (-StdDisableLogging) because warns of type "Could not locate object.d or object.di ..." are visible while they should not

wilzbach:
> Actually the dependence on the ms C library is specified

Hmm fair enough, but shouldn't there be a way to statically bundle it like we can do for Linux? : thinking:

BBasile:
> shouldn't there be a way to statically bundle

The version produced by dmd with `-m64` also has a dependence on a dynamic library (msvcr100.dll) so I'd say that either it's not possible or there are licensing issue. Maybe one of the LDC developer will know better.

release-windows.sh
```diff
@@ -1,26 +1,21 @@
...
 
-# Step 2: Run DMD via wineconsole
-archiveName="dscanner-$VERSION-$OS-$ARCH_SUFFIX.zip"
+# Run LDC with cross-compilation
+archiveName="$BIN_NAME-$VERSION-$OS-$ARCH_SUFFIX.zip"
 echo "Building $archiveName"
 mkdir -p bin
-git describe --tags > bin/githash.txt # no git installed under Wine
-DC="$DIR/dmd2/windows/bin/dmd.exe" wine cmd /C build.bat
+DC=ldmd2 make ldcbuild
```
BBasile: DC=ldmd2 is useless since you `make ldcbuild`.

makefile
```diff
@@ -26,6 +26,9 @@ INCLUDE_PATHS = \
 VERSIONS =
 DEBUG_VERSIONS = -version=dparse_verbose
 DMD_FLAGS = -w -release -O -Jbin -od${OBJ_DIR} -version=StdLoggerDisableWarning
+override DMD_FLAGS += $(DFLAGS)
+override LDC_FLAGS += $(DFLAGS)
+override GDC_FLAGS += $(DFLAGS)
 DMD_TEST_FLAGS = -w -g -Jbin -version=StdLoggerDisableWarning
 override LDC_FLAGS += -O5 -release -oq
```
BBasile: here must be added -version=StdLoggerDisableWarning. for FDC flags too.

LaurentTreguier:
> Hmm so I was testing this in one of the [free VirtualBox for Windows IE](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/), but it complains about the missing `vcruntime140.dll` here:
> 
> image

I tried compiling a binary with LDC and testing it on the Windows 7 image; I got the `missing vcruntime140.dll` error when LDC uses LLD, but not when using Microsoft's linker.
The Windows LDC archive has a [`msvcEnv.bat`](https://github.com/ldc-developers/ldc/blob/master/vcbuild/msvcEnv.bat) script that can be used to load the necessary environment variables (`msvcEnv x64` or `msvcEnv x86`).

> And wine just aborts with "wine: Bad EXE format for ". So I guess we must have missed sth. here.

I got this when trying to launch the x86 binary with wine64, but not with the x64 binary.

BBasile:
@wilzbach  can you address the two small points i've raised after testing the right binaries ?
The dll is not an issue. People will have to install the MS installer for this lib, that's not a problem.
It's very likely that on a real system it's already there (i.e a system used daily since months or years.

So it's almost good now, i don't understand why you stop two meters before the finish line.

PR #747, opened 2019-01-11
SonarQube external issue reporter added

andre2007:
This pull request adds the possibility to set message type and severity. Also an SonarQube external issue reporter is added.

### DIFF
--- a/src/dscanner/analysis/alias_syntax_check.d
+++ b/src/dscanner/analysis/alias_syntax_check.d
@@ -27,10 +27,17 @@ final class AliasSyntaxCheck : BaseAnalyzer
 			return;
 		assert(ad.declaratorIdentifierList.identifiers.length > 0,
 				"Identifier list length is zero, libdparse has a bug");
+
+		MessageDescriptor descr = {
+			message: "Prefer the new \"'alias' identifier '=' type ';'\" syntax"
+				~ " to the  old \"'alias' type identifier ';'\" syntax.",
+			key: KEY,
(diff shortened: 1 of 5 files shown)


BBasile:
So for now the new `metadata()` methods are not used ? I think there's should be something that use them in a test or whatever. Actually i don't even see how they can be used.

andre2007:
Yes you are right. I wanted to start with a rather small pr, but it makes sense to add a use case.

BBasile:
I don't know. Maybe since each check as one or several KEY enum maybe in a first time each check could get as member a something like

```d
struct MessageDescriptor
{
    static string message; // this one is a bit problematic i think, it's mostly used to format...

    static string KEY;
    static MessageType type;
    static MessageSeverity severity;
}
```

Then when emitting a message you save 2 params

andre2007:
@BBasile I think I did not really understood the static field in the struct and how it helps me.
I changed the pr and did a example implementation for alias_syntax_check. 
When the concept is OK and validated, I can change the other checks too.

BBasile:
Indeed the fields must not be static. Sorry i have no idea why i suggested this.

andre2007:
Any feedback ? :)

andre2007:
@Basile-z is the used approach ok for you? If yes I would go on and adapt all analysis modules accordingly.

andre2007:
As Jenkins is now able to read D-Scanner json report files, I close this pr.

PR #752, opened 2019-02-14
update dependencies dparse 0.11.z and containers fixed to 0.8.0-alpha.11

Basile-z:
- dparse 0.11.z is to have dcd-dfmt-dscanner using the same version
- containers fixed version if to prevent https://github.com/dlang-community/containers/issues/136

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -12,11 +12,11 @@
     "StdLoggerDisableWarning"
   ],
   "dependencies" : {
-    "libdparse": "~>0.10.12",
-    "dsymbol" : "~>0.5.5",
+    "libdparse": "~>0.11.0",
+    "dsymbol" : "~>0.6.0",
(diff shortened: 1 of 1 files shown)


LaurentTreguier:
Maybe `dub.json` should target `libdparse`: `~>0.11.2` for consistency with the Git submodule ?

Basile-z:
If it's ok for you since i did this PR for you. I think your vs plugin require the same dparse everywhere, right ?

Basile-z:
well finally it doesn't mater, dub dependency resolver will choose 0.11.2 anyway i think

LaurentTreguier:
Yes it doesn't matter, it was really just to have more or less the same version listed with both dub and git modules, but indeed dub will choose the latest version anyway

Issue #351, opened 2016-05-28
Check resolution suggestions


david7a68:
A number of checks have relatively simple syntactical corrections. It would be beneficial to have recommendations such as: 
`"Prefer the new 'alias identifier = type;' syntax to the  old 'alias type identifier ;' syntax. [recommended_fix, ...]".`

wilzbach:
Do you know about dfix? 
https://github.com/Hackerpilot/dfix

PR #758, opened 2019-04-09
ifelseternary

burner:
A check for patterns

```D
string f;
if(cond) {
   f = "Hello";
} else {
   f = "World";
}
```

to be replaced by 
```D
string f = cond ? "Hello" : "World";
```

### DIFF
--- a/src/dscanner/analysis/ifelseternary.d
+++ b/src/dscanner/analysis/ifelseternary.d
@@ -1,0 +1,212 @@
+//          Copyright Robert burner Schadek 2019.
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file LICENSE_1_0.txt or copy at
+//          http://www.boost.org/LICENSE_1_0.txt)
+
+module dscanner.analysis.ifelseternary;
+
+import std.stdio;
+import dparse.ast;
+import dparse.lexer;
+import dscanner.analysis.base;
+import dscanner.analysis.helpers;
+import dsymbol.scope_ : Scope;
+
+/**
+ * Checks if IfElse Statement can be rewritten as a ternary expression
+ */
+final class IfElseTernary : BaseAnalyzer
+{
+	import std.array : back, empty, popBack;
+
+	private enum IfOrElse
+	{
+		If,
+		Else
+	}
+
+	alias visit = BaseAnalyzer.visit;
+
+	private IfOrElse[] inIfElse;
+	private bool[] isAssign;
+	private string[] thenIdentifier;
+	private string[] elseIdentifier;
+	private int unaryCnt;
+
+	this(string fileName, const(Scope)* sc, bool skipTests = false)
+	{
+		super(fileName, sc, skipTests);
+	}
+
+	override void visit(const IfStatement ifStatement)
+	{
+		import std.format : format;
+
+		// we only need to look at the cases where we have a else
+		if (ifStatement.elseStatement is null)
+		{
+			return;
+		}
+
+		this.inIfElse ~= IfOrElse.If;
+		ifStatement.thenStatement.accept(this);
+		this.inIfElse.popBack();
+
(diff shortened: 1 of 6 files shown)


Basile-z:
Good job, there are quite a bunch of matches in [phobos](https://travis-ci.org/dlang-community/D-Scanner/jobs/517741244#L285) .

WebFreak001:
I think this feature is great but right now it's way too eager.

https://github.com/dlang-community/D-Scanner/pull/758/files#diff-7c757b7c4d7a2962179c7846a5669b88R135
I don't know how to even rewrite this as nice ternary, this if-else is much more readable than anything I can think of.

Can the diagnostics message also suggest the replacement code? It would also make it much easier for text editors to offer automatic replace commands. Of course this would be even easier if dscanner had these fixes built-in like having the source range and a replacement text to insert. - like #351 

Also I think it would be better if it wasn't a warning, but some kind of new diagnostic like "suggestion" which editors wouldn't neccessarily mark as `"warning: your code is bad"`, but rather only show when you hover over the affected source range (or similar)

But I would love to have this feature because sometimes there are cases where a ternary would fit much better.

burner:
> I don't know how to even rewrite this as nice ternary, this if-else is much more readable than anything I can think of.

```D
string person = "unknown";
string hello;
const personUnknown = person == "unknown";
person = personUnknown ? "text" : "bar;
hello = hello.empty && personUnknown ? "foo" : hello;
```

I don't think D-Scanner supports suggestions or automatic rewriting.
I'm open to suggestions, but I don't want to create a new subsystem or anything along that lines to get this in.
And you can always disable this test in your local config.

src/dscanner/analysis/ifelseternary.d
```diff
@@ -0,0 +1,172 @@
+//          Copyright Brian Schott (Hackerpilot) 2014.
```
Hackerpilot: More time-traveling identity theft :trollface:

Hackerpilot:
This case is not detected:

```d
void main(string[] args)
{
    int a;
    if (a == 100)
        a = to!int(args[1]);
    else
        a = 100;
}
```

burner:
@Hackerpilot is there a clean way to get the left side of an assign expression?

burner:
> This case is not detected:
> 
> ```d
> void main(string[] args)
> {
>     int a;
>     if (args.length > 1)
>         a = to!int(args[1]);
>     else
>         a = 100;
> }
> ```

I added a test with that code.

Hackerpilot:
> @Hackerpilot is there a clean way to get the left side of an assign expression?

All of the AST classes have the line `mixin OpEquals;`, which generates an `==` overload that will recursively scan both nodes. This is what I use in the check for both branches of an `if` statement being the same.

You should be able to check that
1. Both branches of the `if` have a single statement.
2. Both statements are assign expressions.
3. The operators for both assign expressions are `=`
4. The left sides of both expressions are equal according to the `==` overload.

Basile-z:
> 3. The operators for both assign expressions are =

The check would be valid for other assign operators as long as both are identical.

burner:
that is not really what I'm looking for.
I'm looking for a way to get the identifiers of the left hand side of the AssignExpressions in the then and else branch.

Basile-z:
if you want the text of the left side text you can use the ast formatter.

burner:
hm, that seems like overkill. I have a implemented a solution in this PR though.

burner:
so what is the way forward?

Basile-z:
Well let's forget the operator for now so this is simple. Memorize the whole LHS expression, and compare (== as explained by HackerPilot)  the one of the if branch with one of the else branch to see if the target is the same. Then to display the LHS exp use the formatter, it's very simple too.

Basile-z:
In the absence of willing to compare the LHS assignment, whatever is its complexity, in the same scope the semantic should be the same.

PR #761, opened 2019-05-14
Fix #760 - ignore deprecated symbols in the undocumented check

### LINKED ISSUE #760: Public documented symbol checker should ignore deprecated symbols
From: https://github.com/dlang/phobos/pull/6998

Example:

```d
deprecated("To be removed in 2.089. Use std.datetime.stopwatch.StopWatch.") @safe auto measureTime(alias func)(){}
```

wilzbach:
Fixes https://github.com/dlang-community/D-Scanner/issues/760


Basile-z:
You can update the dependency to emsi containers to make the CIs all green.

PR #763, opened 2019-05-28
fix #762 - False positive on "unused variable" with side effect

### LINKED ISSUE #762: False positive on "unused variable" with side effect
```D
private:
class Identity { string val; }
Identity[string] arr;

void main ()
{
    arr["Hello"] = new Identity;
    oops();
}

void oops ()
{
    auto v = "Hello" in arr;
    v.val = "Side effect";
}
```

```console
dscanner --styleCheck test.d
test.d(13:10)[warn]: Variable v is never used.
```

This prevents this check from being enabled in DMD, as it relies heavily on this pattern.

Basile-z:
So the problem was a probable confusion between the members of an `AssignExpression`. In the test case, the unary `v.val` is not represented by `assignExpression.expression` but by `assignExpression.ternaryExpression`, which, in the test case, is an `UnaryExpression`. The consequence was that the `interestDepth` counter was not incremented.

### DIFF
--- a/src/dscanner/analysis/unused.d
+++ b/src/dscanner/analysis/unused.d
@@ -166,14 +166,9 @@ final class UnusedVariableCheck : BaseAnalyzer
 
 	override void visit(const AssignExpression assignExp)
 	{
-		if (assignExp.ternaryExpression !is null)
-			assignExp.ternaryExpression.accept(this);
-		if (assignExp.expression !is null)
-		{
-			interestDepth++;
-			assignExp.expression.accept(this);
-			interestDepth--;
-		}
+		interestDepth++;
+		assignExp.accept(this);
+		interestDepth--;
 	}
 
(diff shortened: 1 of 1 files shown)


Basile-z:
@Geod24, is it okay for you ?

Geod24:
I assume you tested it with the test case in https://github.com/dlang-community/D-Scanner/issues/762 ?

Basile-z:
Almost, It's reduced.

Basile-z:
Because the problem was wider and could be encoutered in many other situations, i.e each time a dotvar was assigned.

Geod24:
Then go ahead. If I find other bugs I'll be back anyway :)

Basile-z:
Okay. I just tested the exact code you reported and i can confirm it does not produce a warning anymore.
That being said, the unused_var check is known for being a bit problematic due to the limited scope of the checks so I'm not sure it will work fine in DMD.

PR #764, opened 2019-06-02
Use writeln instead

kubo39:
```console
$ ./bin/dscanner --version # before
v0.7.2%
$ ./bin/dscanner --version # after
v0.7.2
```


Basile-z:
thanks, you can check the other tools if you have some time to. The Possibly have the same problem.

Issue #687, opened 2018-08-22
Add unique error and warning labels


Voultapher:
Right now when running dscanner it reports something like this:
```
main.d(6:25)[warn]: Public declaration 'fill_linear_n' is undocumented.
main.d(10:14)[warn]: Variable `i` initializer is useless because it does not differ from the default value
main.d(18:8)[warn]: Public declaration 'fill_linear_n_lookup_one' is undocumented.
```
Warning 1 and 3 are fairly straight forward, and I can address them by searching `dlang function documentation`. Yet the second warning leaves me hanging telling me what is wrong, yet I have no idea how to fix it, because searching for it is not conclusive.

To help with this issue most linters I know employ a form of short unique label for each error and warning.

Either with an abstract symbol like: `E112` or `W601`
Examples:
- [pycodestlye](http://pycodestyle.pycqa.org/en/latest/intro.html#error-codes)
- [pyflakes](https://lintlyci.github.io/Flake8Rules/)

Another popular option is to use a short unique description like: `no-duplicate-case` or `borrowed_box`
Examples:
- [eslint](https://eslint.org/docs/rules/)
- [clippy](https://rust-lang-nursery.github.io/rust-clippy/master/index.html)

Ideally there is an overview page like in the examples that is SEO optimized and that holds both explanation and possible resolution.

Searching for `dscanner W226` or `dscanner no-duplicate-case` is **much** easier and more likely to lead to the place explaining the rule and how to fix code affected by it.

BBasile:
This was suggested [here too](https://github.com/dlang-community/D-Scanner/issues/613#issuecomment-381963462). Now the point is to have the rationale linked to the warning number. About that I like the doc of pyflakes and Clippy and i'd be in favor of making something similar for D-Scanner. If anyone is talented at making GH pages then start something (unless someone is against the idea, which would be surprising.

Voultapher:
For the record, after some more staring at the warning I figured out how to address it, discouraging an explicit initialisier is questionable imo, even the dlang tour does it that way: https://tour.dlang.org/tour/en/basics/loops.

wilzbach:
I think the two points to take away here are:
- use errors codes (I really like shellcheck's wiki, e.g. https://github.com/koalaman/shellcheck/wiki/SC1004)
- don't activate the "useless initializer" check by default (as it's a very subjective choice and not recommended/activated for DMD/Phobos either)

> discouraging an explicit initialisier is questionable imo

[...]

Issue #613, opened 2018-04-17
Output option name in errors and warnings


Andy-FW:
Sometimes it could be hard to find out what checks one want to disable without studying the comments in the defaultConfig file or grepping the output message within the D-Scanner source code.

Therefore it would be very helpful, to include the name of the triggered check into the output message directly.

In addition, this would make it even easier to grep for specific errors or warnings.

WebFreak001:
you can use the `--report` option to give you the keys if you only need it rarely

BBasile:
The only viable option would be a 2 digits code: `[warn 00]: ...`, otherwise the message would be too polluted.

Andy-FW:
This is now possible using 0.8.0.

PR #769, opened 2019-06-24
Track check name for error messages

belka-ew:
This PR adds `name` key to the `--report` output which contains check name (as specified in the configuration) as suggested here: https://github.com/dlang-community/D-Scanner/issues/613.

Also makes introducing further options, like error codes (https://github.com/dlang-community/D-Scanner/issues/687) easier (would require less changes to the code).

### DIFF
--- a/src/dscanner/analysis/base.d
+++ b/src/dscanner/analysis/base.d
@@ -49,7 +49,7 @@ public:
 
 	protected string getName()
 	{
-		return null;
+		assert(0);
 	}
 
 	Message[] messages()
--- a/src/dscanner/analysis/base.d
+++ b/src/dscanner/analysis/base.d
@@ -18,11 +18,23 @@ struct Message
 	string key;
 	/// Warning message
 	string message;
+	/// Check name
+	string checkName;
 }
 
 enum comparitor = q{ a.line < b.line || (a.line == b.line && a.column < b.column) };
 
 alias MessageSet = RedBlackTree!(Message, comparitor, true);
+
+mixin template AnalyzerInfo(string checkName)
+{
+	enum string name = checkName;
+
+	override protected string getName()
+	{
+		return name;
+	}
+}
 
 abstract class BaseAnalyzer : ASTVisitor
 {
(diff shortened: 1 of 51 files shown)


src/dscanner/analysis/base.d
```diff
@@ -35,6 +47,11 @@ public:
 		_messages = new MessageSet;
 	}
 
+	protected string getName()
+	{
+		return null;
```
Basile-z: I'd rather see an `assert(0);` here.

Basile-z:
I wait a few days before merging. It'd be nice if someone that actually use  the report check this more seriously, e.g @WebFreak001 or @LaurentTreguier

WebFreak001:
serve-d doesn't use the JSON reports (it uses dscanner directly as a library)

I looked at the PR, I don't think the default `assert(false);` implementation makes a lot of sense, instead the whole method should be marked as abstract so it must be implemented.

Basile-z:
I suggested the assert because It's not supposed to happen so it is there to remind any potential people coming with a new check. Alternatively it could be abstract too I think.

Basile-z:
Well, whoever use the D-Scanner reports in a tool, please check this PR.

belka-ew:
There is a `StatsCollector` which inherits from `BaseAnalyzer` as well doesn't have a `name()`. So making `name()` abstract would require to introduce an additional base class for all check classes, or defining `name() { assert(false); }` only in the `StatsCollector`. I shied away from changing the base class everywhere since it requires even more changes. But I an do it if everyone finds it to be better.

belka-ew:
I get no feedback here. Merge?

belka-ew:
@Basile-z would you mind to make a release, please?

Basile-z:
I wanted to update dparse before making a new minor release. D-Scanner already a little week late compared to what DMD supports (change to alias syntax).

belka-ew:
Ok, thanks for the feedback.

Basile-z:
This will come in a few hours, the time to propagate the new lidparse tag (v0.13.0) to the other deps.

Basile-z:
v0.8.0 released now.

Issue #772, opened 2019-07-29
Add rules documentation (level, category and description)


ljmf00:
Code static analyzers should have a level, category and a description associated to each rule to let wrappers know the severity of the error, more about it and categorize to be more easy to search for specific errors (code style, error prone, security issues...)

I'm open to contribute and implement this, but before making a pull request, should discuss about the type of documentation that is going to be used. I suggest to follow [this](https://support.codacy.com/hc/en-us/articles/207994335-Code-Patterns) documentation. A `json` or `xml` and a specification for that file should be enough for wrappers to know how the tool work.

PR #775, opened 2019-09-06
Added reportFormat sonarQubeGenericIssueData

andre2007:
This feature pr will allow to directly produce a JSON output in [SonarQube Generic Issue Data format](https://docs.sonarqube.org/latest/analysis/generic-issue/).

Therefore you can analyze your D coding on SonarQube 7.x without the (unmaintained) d plugin for SonarQube but directly by using the generic issue format of SonarQube.

By default dscanner will output the dscanner json format. By using new option ```--reportFormat sonarQubeGenericIssueData``` the SonarQube Generic Issue Data format is produced.


src/dscanner/reports.d
```diff
@@ -0,0 +1,160 @@
...
+		// dfmt on
+		return issue;
+	}
+
+	private Location getPrimaryLocation(Message message)
+	{
+		return Location(message.message, message.fileName, TextRange(message.line));
+	}
+
+	private string getSeverity(string key)
```
WebFreak001: could this be made `public static` or possibly be moved outside the class? I think this might be useful in other applications like editors directly integrating DScanner. It should probably be documented how the naming affects the severity then too.

I think the method could also take in a `const ref Message message` argument so in case we add some severity/type enum to the message, it can simply translate to the SonarQube enums and otherwise now just only use `.key`.

src/dscanner/reports.d
```diff
@@ -0,0 +1,160 @@
...
+			{
+				case "style":
+					return Severity.minor;
+				default:
+					return Severity.major;
+			}
+		}
+	}
+
+	private string getType(string key)
```
WebFreak001: ditto

andre2007:
@WebFreak001 I started a pr some time ago implementing your proposal 
 https://github.com/dlang-community/D-Scanner/pull/747 but at the end didn't got any feedback anymore. 
It implies a huge amount of work to enable all checks with type and severity, without feedback I stopped the work.

The two methods are just a workaround. My gut feeling is, they don't make sense as a general library solution and they will get obsolete when message type and severity is added to all check modules.

WebFreak001:
oh yes sorry. I think the PR itself was too big. A big change like adding descriptions to all stuff should be a PR with just that content and nothing else (adding SonarQube on it just makes it more uncomfortable for people to look at who don't really know it)

Also maintainability is a problem with that, it should definitely have separate files which are read for descriptions. Maybe a JSON containing HTML code or separate HTML files for each issue? Best to contribute on #772 for that.

I think having a separate module taking in a message ID and giving you these values at runtime would be a good idea as it's the most configurable. But I guess that would make most sense to discuss in that issue.

If we can get these changes in as separate PR before this PR then we can just directly use these attributes too. So if you think this can be done in a short time frame we can do that, otherwise just merge this with the private methods.

src/dscanner/reports.d
```diff
@@ -0,0 +1,160 @@
...
+
+	struct TextRange
+	{
+		// SonarQube Generic Issue only allows specifying start line only
+		// or the complete range, for which start and end has to differ
+		// D-Scanner does not provide the complete range info
+		long startLine;
+	}
+
+	private Issue[] _issues;
```
WebFreak001: I think this would be better as `Appender!(Issue[])`

src/dscanner/reports.d
```diff
@@ -0,0 +1,160 @@
...
+
+	string getContent()
+	{
+		JSONValue result = [
+			"issues" : JSONValue(_issues.map!(e => toJson(e)).array)
+		];
+		return result.toString();
+	}
+
+	private JSONValue toJson(Issue issue)
```
WebFreak001: maybe all the private methods here should be made `static`?

src/dscanner/main.d
```diff
@@ -91,6 +92,7 @@ else
 				"declaration|d", &symbolName,
 				"config", &configLocation,
 				"report", &report,
+				"reportFormat", &reportFormat,
```
WebFreak001: setting this should implicitly set `report` to `true`.

src/dscanner/main.d
```diff
@@ -237,7 +239,12 @@ else
 		if (skipTests)
 			config.enabled2SkipTests;
 		if (report)
-			generateReport(expandArgs(args), config, cache, moduleCache);
+		{
+			if (reportFormat == "sonarQubeGenericIssueData")
```
WebFreak001: maybe this should be made a switch-case so a warning message can be logged on default (with fallthrough)

switch-case also implies that there may be more formats to be added in the future.

src/dscanner/reports.d
andre2007: fixed

src/dscanner/reports.d
andre2007: fixed

src/dscanner/main.d
andre2007: yes, really make sense. fixed

src/dscanner/main.d
andre2007: fixed

src/dscanner/reports.d
WebFreak001: @andre2007 fyi you can click "Resolve conversation" to hide the outdated conversations

src/dscanner/reports.d
```diff
@@ -0,0 +1,160 @@
...
+
+	struct TextRange
+	{
+		// SonarQube Generic Issue only allows specifying start line only
+		// or the complete range, for which start and end has to differ
+		// D-Scanner does not provide the complete range info
+		long startLine;
+	}
+
+	private Appender!(Issue[]) _issues;
```
WebFreak001: I think this is missing initialization? `_issues = appender!(Issue[]);`

I don't think it's specified what will happen if you don't initialize it, but `appender` is documented to be a convenience function returning an instance.

The constructor of Appender is doing a bit of work so I think it should be called in the constructor in the class: https://github.com/dlang/phobos/blob/master/std/array.d#L3239

src/dscanner/main.d
```diff
@@ -155,6 +157,9 @@ else
 	if (absImportPaths.length)
 		moduleCache.addImportPaths(absImportPaths);
 
+	if (reportFormat != "")
```
WebFreak001: I think this comparision can have issues if reportFormat was initialized to `null`. I would do it like in the if check above and use `if (reportFormat.length)` instead

src/dscanner/analysis/run.d
```diff
@@ -232,9 +260,15 @@ const(Module) parseModule(string fileName, ubyte[] code, RollbackAllocator* p,
...
-	return dparse.parser.parseModule(tokens, fileName, p,
-		report ? toDelegate(&messageFunctionJSON) : writeMessages,
-		errorCount, warningCount);
+
+	MessageDelegate messageWriter;
+
+	if (dlgMessage !is null)
+		messageWriter = dlgMessage;
+	else
+		messageWriter = report ? toDelegate(&messageFunctionJSON) : writeMessages;
```
WebFreak001: I think having a new overload to parseModule with a non-optional MessageDelagate argument and no `bool report` nor `string errorFormat` argument would make the API a bit more cleaner. The MessageDelegate would then also be before the int pointers which matches the underlying API too.

The one with the old overload could then call this new overload and it would no longer create a new delegate in this method.

andre2007: fixed

src/dscanner/main.d
andre2007: fixed

src/dscanner/reports.d
andre2007: fixed

src/dscanner/analysis/run.d
```diff
@@ -180,6 +181,32 @@ void generateReport(string[] fileNames, const StaticAnalysisConfig config,
...
+
+	foreach (fileName; fileNames)
+	{
+		auto code = readFile(fileName);
+		// Skip files that could not be read and continue with the rest
+		if (code.length == 0)
+			continue;
+		RollbackAllocator r;
+		const(Token)[] tokens;
+		const Module m = parseModule(fileName, code, &r, cache, tokens, writeMessages, &lineOfCodeCount, null, null);
```
WebFreak001: the `lineOfCodeCount` argument is not used anywhere and can be replaced with `null`

src/dscanner/main.d
```diff
@@ -237,7 +242,21 @@ else
...
+			{
+				case "":
+				case "dscanner":
+					generateReport(expandArgs(args), config, cache, moduleCache);
+					break;
+				case "sonarQubeGenericIssueData":
+					generateSonarQubeGenericIssueDataReport(expandArgs(args), config, cache, moduleCache);
+					break;
+				default:
+					stderr.writeln("Unknown report format specified, using default format");
```
WebFreak001: maybe to avoid code duplication you could write
```d
default:
	stderr.writeln("Unknown report format specified, using dscanner format");
	goto case;
case "":
case "dscanner":
	...
```

Also it should log "using dscanner format"

andre2007: fixed

src/dscanner/analysis/run.d
andre2007: fixed

andre2007:
@WebFreak001 thanks a lot for looking at this pr and your time.

andre2007:
@WebFreak001 I added a short description to README.md, please check.

Hackerpilot:
Does the bot not look at the auto-merge tag? I just noticed that https://github.com/dlang-community/D-Scanner/pull/773 hasn't been merged either.

PR #776, opened 2019-09-23
Support checkName in errorFormat

belka-ew:
Adds support for the check name in `--errorFormat` and updates the help message a bit.

### DIFF
--- a/src/dscanner/main.d
+++ b/src/dscanner/main.d
@@ -376,6 +376,8 @@ Options:
     --errorFormat|f <pattern>
         Format errors produced by the style/syntax checkers. The default
         value for the pattern is: "%2$s".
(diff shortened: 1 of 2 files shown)


src/dscanner/main.d
```diff
@@ -376,6 +376,8 @@ Options:
     --errorFormat|f <pattern>
         Format errors produced by the style/syntax checkers. The default
         value for the pattern is: "%2$s".
```
ZombineDev: Maybe I'm missing something, but shouldn't this text say that the default value for the pattern is:
`"{filepath}({line}:{column})[{type}]: {message}"`?

andre2007:
Does someone know, when the next release of D-Scanner is planned?

PR #787, opened 2020-01-04
Fix empty error messages in issue 786

MoonlightSentinel:
The message was assigned to Message.key and hence not printed.

Note that this does _not_ fix the issue itself, just the hidden messages.


WebFreak001:
uhh not quite sure if that merge branch into master thing was needed, my github showed me that button so I thought your branch was out of date, but anyway it is merged now (with 3 commits lol)

MoonlightSentinel:
It was based on the current master ... but no (real) harm in unnecessary commits

PR #790, opened 2020-01-26
Fix #788: Enum value used inside __traits(...) flagged as unused

### LINKED ISSUE #788: Enum value used inside __traits(...) flagged as unused
Reduced test case:
```
void setMsgSendSymbol()
{
    enum fieldName = "abc";
    __traits(hasMember, S, fieldName);
}
```

```
style.d(3:10)[warn]: Variable fieldName is never used.
```

This warning is e.g. not issued for:
```
enum fieldName = "abc";
static assert(!__traits(getMember, S, fieldName));
```

MoonlightSentinel:


MoonlightSentinel:
Is this good to go?

PR #800, opened 2020-03-21
Exclude `.git` from search paths.

patha454:
Currently dscanner searches for files in the target directory ending in `.d` or `.di`.

EDIT: This PR at first referred to just the `.git` directory. I've since widened the scope to cover all hidden directories. 

Git has some hidden files in `.git`, which log commits to files. These files share the same name as the tracked file: For example `.git/logs/refs/heads/somedfile.d` actually contains Git data, not D code.

A naive call to dscanner using `dscanner --syntaxCheck .` or `dscanner --styleCheck .` (therefore also `dub lint...`), inside a D project whichs is also a Git repository, will attempt to check these Git files. The result is spurious errors, such as:

```sh
./.git/logs/refs/heads/yourfile.d(1:1)[error]:
./.git/logs/refs/heads/yourfile.d(1:118)[warn]: Line is longer than 120 characters
```

I believe Git is a common enough tool that we're justified to handle this case inside D scanner, rather than asking the user to write configuration files with dodge the `.git` folder.

Alternatively we could add an `--ignore $PATH` command line argument to filter out paths eaisly.

### DIFF
--- a/src/dscanner/utils.d
+++ b/src/dscanner/utils.d
@@ -84,7 +84,7 @@ string[] expandArgs(string[] args)
 {
 	import std.file : isFile, FileException, dirEntries, SpanMode;
 	import std.algorithm.iteration : map;
-	import std.algorithm.searching : endsWith;
+	import std.algorithm.searching : endsWith, find;
 
 	// isFile can throw if it's a broken symlink.
 	bool isFileSafe(T)(T a)
@@ -105,7 +105,7 @@ string[] expandArgs(string[] args)
 		else
 			foreach (item; dirEntries(arg, SpanMode.breadth).map!(a => a.name))
 			{
-				if (isFileSafe(item) && (item.endsWith(`.d`) || item.endsWith(`.di`)))
+				if (isFileSafe(item) && (item.endsWith(`.d`) || item.endsWith(`.di`)) && !item.find('.git'))
 					rVal ~= item;
 				else
 					continue;
--- a/src/dscanner/utils.d
+++ b/src/dscanner/utils.d
@@ -4,7 +4,6 @@ import std.array : appender, uninitializedArray;
 import std.stdio : stdin, stderr, File;
 import std.conv : to;
 import std.encoding : BOM, BOMSeq, EncodingException, getBOM;
-import std.path : dirSeparator;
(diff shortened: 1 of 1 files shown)


rikkimax:
Please change this to all directories beginning with '.'.

There is no reason for hidden directories such as these to be searched by default.

patha454:
> 
> 
> Please change this to all directories beginning with `.`.
> 
> There is no reason for hidden directories such as these to be searched by default.

Agreed. I'll get onto this.

Hackerpilot:
The CI passes for this pull request There was some intermittent failure on Travis, so I just had to restart the builds.

src/dscanner/utils.d
```diff
@@ -105,7 +106,7 @@ string[] expandArgs(string[] args)
 		else
 			foreach (item; dirEntries(arg, SpanMode.breadth).map!(a => a.name))
 			{
-				if (isFileSafe(item) && (item.endsWith(`.d`) || item.endsWith(`.di`)))
+				if (isFileSafe(item) && (item.endsWith(`.d`) || item.endsWith(`.di`)) && !item.canFind(dirSeparator ~ '.'))
```
WebFreak001: wouldn't the `dirSeparator ~ '.'` allocate a lot of garbage? I think it should be put in a constant outside the loops.

I can't test if it does though because run.dlang.io doesn't support emitting the profile=gc log and ldc doesn't support it either and dmd isn't downloadable here.

WebFreak001: ok nvm could test now and turns out it doesn't even allocate with dirSeparator

Hackerpilot: I would be very surprised if the compiler couldn't constant fold that.

patha454: Aye. `dirSeperator` is an enum constant. `"."` is a compile time constant. I've never read the compiler source, but I'd assume it's a compile time constant and treated as such.

Hackerpilot:
@WebFreak001 Any more comments on this?

WebFreak001:
LGTM otherwise

Hackerpilot:
I guess we don't have @dlang-bot working here.

PR #805, opened 2020-04-27
Use const instead of immutable for compatibility with new libdparse

Hackerpilot:
Looks like the libdparse token changes weren't as backwards-compatible as we hoped. See https://github.com/dlang-community/D-Scanner/pull/799#issuecomment-617104293

### DIFF
--- a/src/dscanner/analysis/auto_function.d
+++ b/src/dscanner/analysis/auto_function.d
@@ -78,7 +78,7 @@ public:
 			if (!p)
 				return;
 
-			immutable token = p.primary;
+			const token = p.primary;
 			if (token.type == tok!"false")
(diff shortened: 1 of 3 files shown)


WebFreak001:
hm I'm not quite sure how these changes are related to the token structure changes. Did including the const arrays make it impossible to be immutable?

In this case would changing
```
    const(typeof(this))[] leadingTrivia;
    /// ditto
    const(typeof(this))[] trailingTrivia;
```
in libdparse to
```
    immutable(typeof(this))[] leadingTrivia;
    /// ditto
    immutable(typeof(this))[] trailingTrivia;
```
be the needed fix? The appender data is dup'd anyway, it could probably be changed to idup without issues.

PR #807, opened 2020-06-08
check libdparse & dsymbol with min/max dub version

WebFreak001:
analogous to https://github.com/dlang-community/dfmt/pull/491 this will run the dub tests first with minimum libdparse version, then with maximum libdparse version, then with minimum dsymbol version and with maximum dsymbol version.

This makes version ranges of dependencies easier to test.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -3,7 +3,14 @@
 set -e
 
 if [[ $BUILD == dub ]]; then
-    dub test
+    if [[ -n $LIBDPARSE_VERSION ]]; then
+        rdmd ./d-test-utils/test_with_package.d $LIBDPARSE_VERSION libdparse -- dub test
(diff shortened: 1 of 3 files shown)


.travis.sh
```diff
@@ -3,7 +3,14 @@
 set -e
 
 if [[ $BUILD == dub ]]; then
-    dub test
+    if [[ -n $LIBDPARSE_VERSION ]]; then
+        rdmd ./d-test-utils/test_with_package.d $LIBDPARSE_VERSION libdparse -- dub test
+    elif if [[ -n $DSYMBOL_VERSION ]]; then
```
Hackerpilot, line 8: I'm pretty sure the `if` should not be here.

WebFreak001: fixed

PR #812, opened 2020-07-17
Document some of the undocumented command-line switches in --help output.

Hackerpilot:


src/dscanner/main.d
```diff
@@ -348,6 +348,11 @@ Options:
         Prints the number of tokens in the given source files. If no files are
         specified, input is read from stdin.
 
+    --tokenDump <file>
+        Dump token information from the lexer. This option is mostly useful for
+        developing D-Scanner or its supporting libraries itself. You probably
+        dont't want to use this, and this feature may be removed in the future.
+
```
WebFreak001, line 355: I don't see why we should include this in the documentation at all

src/dscanner/main.d
```diff
@@ -361,6 +366,10 @@ Options:
     --recursiveImports <file>
         Similar to "--imports", but lists imports of imports recursively.
 
+    -I <directory>
+        Specify that imported the given directory should be searched for
+        imported modules.
+
```
WebFreak001, line 372: I think this is useful enough to add to README

Hackerpilot: It's already there. https://github.com/dlang-community/D-Scanner#import-listing

src/dscanner/main.d
Hackerpilot: I debated leaving this out, but it may be useful to someone. I figured that if it's in the tool (and not one of the dumb Easter eggs) it's better to document it.

PR #813, opened 2020-07-22
Bump libdparse to up to 0.15.x

WebFreak001:


WebFreak001:
would merge and release once all tests passed.

Only a single change in behavior in the release, but needed for being able to support the latest libdparse.

PR #818, opened 2020-08-20
Update libdparse to fix Issue 817

### LINKED ISSUE #817: If the UDA is used in a module, strange errors occur
Using UDAs in modules will make an incorrect error report:
```d
@("test")
module main; // src\main.d(2:1)[error]: Declaration expected

void main()
{
}
```

It's a pretty special case, but the specification allows for it.
https://dlang.org/spec/module.html#module_declaration
```
ModuleAttribute:
    DeprecatedAttribute
    UserDefinedAttribute
```

Hackerpilot:
Fixes #817.


Hackerpilot:
After doing a lot of troubleshooting with various dub dependencies, the build is green again.

PR #819, opened 2020-08-21
Add unused_result check

CyberShadow:
```d
/**
 * Checks for function call statements which call non-void functions.
 *
 * In case the function returns a value indicating success/failure,
 * ignoring this return value and continuing execution can lead to
 * undesired results.
 *
 * When the return value is intentionally discarded, `cast(void)` can
 * be prepended to silence the check.
 */
```

Unfortunately gimped by Dscanner design limitations: it only works for global functions *within the same module*.


CyberShadow:
Thanks! :)

PR #830, opened 2021-05-02
fix if_constraints_indent with broken/null tokens Fix #829

### LINKED ISSUE #829: memory leak in if_constraints_indent with empty char (broken token)
https://github.com/dlang-community/D-Scanner/blob/2963358eb4a24064b0893493684d4075361297eb/src/dscanner/analysis/if_constraints_indent.d#L30

because t.line is unsigned, the subtraction makes this effectively an infinite loop, constantly appending more memory.

The issue may come from the caller, but it's using a directly passed tokens array from the analyze function, so this may come from apps using D-Scanner as a library as well.

See https://github.com/Pure-D/serve-d/issues/135

WebFreak001:
Subtraction from t.line (which is unsigned) caused the while loop
to run nearly infinitely before, starving the system of memory as it
was allocating memory.

fixes #829

### DIFF
--- a/src/dscanner/analysis/if_constraints_indent.d
+++ b/src/dscanner/analysis/if_constraints_indent.d
@@ -24,12 +24,19 @@ final class IfConstraintsIndentCheck : BaseAnalyzer
 	{
 		super(fileName, null, skipTests);
 
+		// convert tokens to a list of token starting positions per line
+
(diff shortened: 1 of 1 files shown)


WebFreak001 (commented):
self review looks ok (could maybe rewrite first comment a little clearer)

PR #834, opened 2021-06-02
Bump liddoc and libdparse versions

RazvanN7:
I'm trying to make dscanner use the latest versions of libddoc and libdparse to be able to recognize the AliasAssign declarations. This will unblock multiple phobos PRs such as [1]. Can someone, please provide some assistance as to what is necessary to make the phobos testing pipeline recognize the latest version of dscanner (the one with the latest versions of libddoc and libdparse).

Thanks!

[1] https://github.com/dlang/phobos/pull/8040

PR #835, opened 2021-06-05
actually bump libdparse now

WebFreak001:
forgot it last PR (#834)

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -12,7 +12,7 @@
     "StdLoggerDisableWarning"
   ],
   "dependencies" : {
-    "libdparse": ">=0.17.0 <0.18.0",
+    "libdparse": ">=0.17.0 <0.19.0",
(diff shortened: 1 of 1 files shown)


PetarKirov:
I think you also need to update the git submodule commit hash

WebFreak001:
git submodule already points to the later version because of last PR (that's why this is the corrective PR)

WebFreak001:
> ```
> Unresolvable dependencies to package libdparse:
>  dscanner 0.11.1+commit.3.g18867d4 depends on libdparse >=0.17.0 <0.19.0
>  dsymbol 0.11.3 depends on libdparse >=0.14.0 <0.18.0
>  libddoc 0.8.0 depends on libdparse >=0.13.0 <1.0.0
> ```

doesn't 0.17.x fit those restrictions? why would that be unresolvable?

RazvanN7:
I think that this can be closed now.

PR #836, opened 2021-06-12
WIP: Support for exporting CodeClimate compatible reports.

nazriel:


andre2007:
SonarQube has a report format "Generic Issue Data". This the reason for the SonarQube class name. I assume you copied the class name, but likely "generic issue data" does not make sense for your new class.

rikkimax:
Based upon the old CI result, this might be mergeable.

But it'll need a push to trigger the current CI to verify that (I think).

WebFreak001:
on master we now have diagnostic end locations / ranges, instead of just line and column, as well as supplemental diagnostic locations. You might want to integrate this in your report format.

nazriel:
Hi guys, sorry for the troubles. I forgot about this MR.
I will close it and redo with newest code-base and your ideas

Issue #827, opened 2021-02-02
Assertion failure in `dsymbol.modulecache`


chances:
See this CI run: https://github.com/chances/teraflop-d/pull/11/checks?check_run_id=1817643809#step:8:41

I'm not sure if this is a problem with dsymbol directly, see 

```
core.exception.AssertError@../../../.dub/packages/dsymbol-0.11.3/dsymbol/src/dsymbol/modulecache.d(261): Assertion failure
```

https://github.com/dlang-community/dsymbol/blob/b8920c2a4593e24d1ced91f8f0bce9e50f598e4d/src/dsymbol/modulecache.d#L261

RazvanN7:
This is currently blocking the update of the phobos pipeline: https://buildkite.com/dlang/phobos/builds/5801#2449d9be-88e2-4a10-88ee-2a01dad6d71a

PR #837, opened 2021-07-28
Upgrade libdparse version

RazvanN7:
Contains the fix of a bug that is causing memory corruption + unblocks https://github.com/dlang/dmd/pull/12898

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -12,7 +12,7 @@
     "StdLoggerDisableWarning"
   ],
   "dependencies" : {
-    "libdparse": ">=0.17.0 <0.18.0",
+    "libdparse": ">=0.17.0 <1.0.0",
     "dsymbol" : ">=0.11.3 <0.12.0",
     "inifiled" : "~>1.3.1",
     "emsi_containers" : "~>0.8.0",


dub.json
```diff
@@ -12,7 +12,7 @@
     "StdLoggerDisableWarning"
   ],
   "dependencies" : {
-    "libdparse": ">=0.17.0 <0.18.0",
+    "libdparse": ">=0.17.0 <1.0.0",
```
Geod24, line 15:
```suggestion
    "libdparse": ">=0.17.0 <0.18",
```

But I don't see why we're not using path dependencies here ??

RazvanN7: I saw that @WebFreak001 used 1.0.0 here [1] and @Hackerpilot validated it, so I went with this.

[1] https://github.com/dlang-community/libddoc/pull/54/files#diff-1b617af56e869773e32ce59329f25b2db5d8b3c5151519919c7c10266b4d78fcR23

RazvanN7: > But I don't see why we're not using path dependencies here ??

Beats me.

WebFreak001: in ddoc the libdparse version is not as rightly coupled, it's only needed for tokenizing - I think using 1.0.0 here is a bad idea because D-Scanner is much more tightly coupled with the API

RazvanN7:
It turns out D-Scanner still has issues [1]. The assert is in dsymbol since 2015 so it might be that dscanner is at fault here.

[1] https://buildkite.com/dlang/phobos/builds/5801#2449d9be-88e2-4a10-88ee-2a01dad6d71a

RazvanN7:
It seems that this is a known issue: https://github.com/dlang-community/D-Scanner/issues/827

PR #838, opened 2021-07-29
Upgrade dsymbol version

RazvanN7:
When trying to upgrade the dscanner version of phobos an assertion failure occurs in dsymbol [1]. Trying to upgrade to the latest version of dsymbol (which also uses the newer version of libdpare) to see if it fixes the issue.

[1] https://buildkite.com/dlang/phobos/builds/5799#95955563-1eb8-472c-8c55-bc9217bc4b16


RazvanN7:
cc @Geod24

PR #840, opened 2021-08-15
Use @safe when possible

nordlow:


src/dscanner/analysis/auto_ref_assignment.d
```diff
@@ -9,6 +9,8 @@ import dparse.lexer;
 import dparse.ast;
 import dscanner.analysis.base;
 
+
+
```
RazvanN7, line 13: Maybe get rid of these unrelated changes?

nordlow: Done.

src/dscanner/main.d
```diff
@@ -37,12 +37,14 @@ import inifiled;
...
+@safe:
+
 version (unittest)
-	void main()
+void main() @trusted
 {
 }
 else
-	int main(string[] args)
+int main(string[] args) @trusted
```
RazvanN7, line 47: What is the point of this change?

RazvanN7 (commented):
Unless I am missing something, this doesn't change the allocator at all. It simply adds safe and trusted.

nordlow:
> Unless I am missing something, this doesn't change the allocator at all. It simply adds safe and trusted.

You're correct. I'll update the title.

src/dscanner/analysis/base.d
```diff
@@ -81,7 +81,7 @@ protected:
 		override void visit(const T structDec)
 		{
 			inAggregate = true;
-			structDec.accept(this);
+			() @trusted { structDec.accept(this); } ();
```
WebFreak001, line 84: I don't think this accept call can generally be `@trust`ed, this should probably stay `@system`

src/dscanner/main.d
WebFreak001: shouldn't main be `@safe` instead of `@trusted`?

PR #852, opened 2022-01-21
add cyclomatic complexity check

### LINKED ISSUE #684: Add a cyclomatic complexity check

WebFreak001:
Fix #684

### DIFF
--- a/src/dscanner/analysis/config.d
+++ b/src/dscanner/analysis/config.d
@@ -209,6 +209,12 @@ struct StaticAnalysisConfig
 	@INI("Check for unused function return values")
 	string unused_result = Check.enabled;
 
+	@INI("Enable cyclomatic complexity check")
+	string cyclomatic_complexity = Check.disabled;
+
+	@INI("Maximum cyclomatic complexity after which to issue warnings")
+	int max_cyclomatic_complexity = 50;
+
 	@INI("Module-specific filters")
 	ModuleFilters filters;
 }
--- a/src/dscanner/analysis/cyclomatic_complexity.d
+++ b/src/dscanner/analysis/cyclomatic_complexity.d
@@ -1,0 +1,261 @@
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file LICENSE_1_0.txt or copy at
+//          http://www.boost.org/LICENSE_1_0.txt)
+
+module dscanner.analysis.cyclomatic_complexity;
+
+import dparse.ast;
+import dparse.lexer;
+import dsymbol.scope_ : Scope;
+import dscanner.analysis.base;
+import dscanner.analysis.helpers;
+
+import std.format;
+
+/// Implements a basic cyclomatic complexity algorithm using the AST.
+///
+/// Issues a warning on functions whenever the cyclomatic complexity of them
+/// passed over a configurable threshold.
+///
+/// The complexity score starts at 1 and is increased each time on
+/// - `if`
+/// - switch `case`
+/// - any loop
+/// - `&&`
+/// - `||`
+/// - `?:` (ternary operator)
+/// - `throw`
+/// - `catch`
+/// - `return`
+/// - `break` (unless in case)
+/// - `continue`
+/// - `goto`
+/// - function literals
+///
+/// See: https://en.wikipedia.org/wiki/Cyclomatic_complexity
+/// Rules based on http://cyvis.sourceforge.net/cyclomatic_complexity.html
+/// and https://github.com/fzipp/gocyclo
+final class CyclomaticComplexityCheck : BaseAnalyzer
+{
+	/// Message key emitted when the threshold is reached
+	enum string KEY = "dscanner.metric.cyclomatic_complexity";
+	/// Human readable message emitted when the threshold is reached
+	enum string MESSAGE = "Cyclomatic complexity of this function is %s.";
+	mixin AnalyzerInfo!"cyclomatic_complexity";
+
+	/// Maximum cyclomatic complexity. Once the cyclomatic complexity is greater
+	/// than this threshold, a warning is issued.
+	///
+	/// By default 50 is used as threshold, which is considered almost
+	/// unmaintainable / untestable.
+	///
+	/// For clean development a threshold like 20 can be used instead.
+	int maxCyclomaticComplexity;
+
+	///
+	this(string fileName, const(Scope)* sc, bool skipTests = false,
+		int maxCyclomaticComplexity = 50)
+	{
+		super(fileName, sc, skipTests);
+		this.maxCyclomaticComplexity = maxCyclomaticComplexity;
+	}
+
+	mixin VisitComplex!IfStatement;
+	mixin VisitComplex!CaseStatement;
+	mixin VisitComplex!CaseRangeStatement;
+	mixin VisitLoop!DoStatement;
+	mixin VisitLoop!WhileStatement;
+	mixin VisitLoop!ForStatement;
+	mixin VisitLoop!ForeachStatement;
+	mixin VisitComplex!AndAndExpression;
+	mixin VisitComplex!OrOrExpression;
+	mixin VisitComplex!TernaryExpression;
+	mixin VisitComplex!ThrowStatement;
+	mixin VisitComplex!Catch;
+	mixin VisitComplex!LastCatch;
+	mixin VisitComplex!ReturnStatement;
+	mixin VisitComplex!FunctionLiteralExpression;
+	mixin VisitComplex!GotoStatement;
+	mixin VisitComplex!ContinueStatement;
+
+	override void visit(const SwitchStatement n)
+	{
+		inLoop.assumeSafeAppend ~= false;
+		scope (exit)
+			inLoop.length--;
+		n.accept(this);
+	}
+
+	override void visit(const BreakStatement b)
+	{
+		if (b.label !is Token.init || inLoop[$ - 1])
+			complexityStack[$ - 1]++;
+	}
+
+	override void visit(const FunctionDeclaration fun)
+	{
+		if (!fun.functionBody)
+			return;
+
+		complexityStack.assumeSafeAppend ~= 1;
+		inLoop.assumeSafeAppend ~= false;
+		scope (exit)
+		{
+			complexityStack.length--;
+			inLoop.length--;
+		}
+		fun.functionBody.accept(this);
+		testComplexity(fun.name.line, fun.name.column);
+	}
+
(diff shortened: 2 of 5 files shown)


CyberShadow:
Have you tried it out on Phobos (or other large packages)?

src/dscanner/analysis/cyclomatic_complexity.d
```diff
@@ -0,0 +1,260 @@
...
+/// The complexity score starts at 1 and is increased each time on
+/// - `if`
+/// - switch `case`
+/// - any loop
+/// - `&&`
+/// - `||`
+/// - `?:` (ternary operator)
+/// - `throw`
+/// - `catch`
+/// - `return`
```
CyberShadow, line 29: Giving `return` a negative cost means that pyramid-of-doom style has a better score than early-return style :) But I guess that would be offset by some maximum-nesting check or something.

src/dscanner/analysis/cyclomatic_complexity.d
```diff
@@ -0,0 +1,260 @@
...
+
+	///
+	this(string fileName, const(Scope)* sc, bool skipTests = false,
+		int maxCyclomaticComplexity = 50)
+	{
+		super(fileName, sc, skipTests);
+		this.maxCyclomaticComplexity = maxCyclomaticComplexity;
+	}
+
+	// based on http://cyvis.sourceforge.net/cyclomatic_complexity.html
```
CyberShadow, line 61: :+1: Maybe put the source of the definition used here somewhere more prominent like the module header.

src/dscanner/analysis/config.d
```diff
@@ -209,6 +209,12 @@ struct StaticAnalysisConfig
 	@INI("Check for unused function return values")
 	string unused_result = Check.enabled;
 
+	@INI("Enable cyclomatic complexity check")
+	string cyclomatic_complexity = Check.enabled;
+
+	@INI("Maximum cyclomatic complexity after which to issue warnings")
+	int max_cyclomatic_complexity = 50;
```
CyberShadow, line 216: I wonder if there should be a separate class of warnings to distinguish "you might have a bug here" from "you need to change your code due to our team's policy / to make sure other people can easily understand/maintain it".

CyberShadow (approved):
LGTM provided that the CI passes.

WebFreak001:
I only run this so far on dscanner itself, where there are 2 cases where there were these cases I agree with:
```
src/dscanner/analysis/useless_initializer.d(136:16)[warn]: Cyclomatic complexity of this function is 57.
src/dscanner/analysis/run.d(363:12)[warn]: Cyclomatic complexity of this function is 113.
src/dscanner/main.d(45:6)[warn]: Cyclomatic complexity of this function is 68.
```

this is definitely complex code that should be broken down:
https://github.com/dlang-community/D-Scanner/blob/7c54e445c6efc7a521bd5667759061c35fc4013c/src/dscanner/analysis/useless_initializer.d#L136

this should probably be replaced with a loop and some kind of standard constructor customization (config maybe)
https://github.com/dlang-community/D-Scanner/blob/7c54e445c6efc7a521bd5667759061c35fc4013c/src/dscanner/analysis/run.d#L362

I think breaking down main would be nice too, but maybe this is something you would wanna ignore:
https://github.com/dlang-community/D-Scanner/blob/master/src/dscanner/main.d#L45

WebFreak001:
introduced phobos warnings:
```
phobos/std/algorithm/searching.d(1899:4)[warn]: Cyclomatic complexity of this function is 61.
phobos/std/datetime/systime.d(10305:9)[warn]: Cyclomatic complexity of this function is 59.
phobos/std/experimental/logger/core.d(2684:7)[warn]: Cyclomatic complexity of this function is 109.
phobos/std/experimental/checkedint.d(2543:1)[warn]: Cyclomatic complexity of this function is 69.
phobos/std/format/internal/floats.d(589:14)[warn]: Cyclomatic complexity of this function is 85.
phobos/std/format/internal/write.d(198:14)[warn]: Cyclomatic complexity of this function is 76.
phobos/std/format/internal/write.d(3375:6)[warn]: Cyclomatic complexity of this function is 56.
phobos/std/internal/math/gammafunction.d(764:6)[warn]: Cyclomatic complexity of this function is 72.
phobos/std/internal/unicode_tables.d(3778:6)[warn]: Cyclomatic complexity of this function is 60.
phobos/std/math/exponential.d(429:17)[warn]: Cyclomatic complexity of this function is 109.
phobos/std/math/operations.d(1492:5)[warn]: Cyclomatic complexity of this function is 55.
phobos/std/net/isemail.d(60:13)[warn]: Cyclomatic complexity of this function is 184.
phobos/std/net/isemail.d(1350:8)[warn]: Cyclomatic complexity of this function is 125.
phobos/std/numeric.d(1157:20)[warn]: Cyclomatic complexity of this function is 68.
phobos/std/regex/internal/backtracking.d(321:9)[warn]: Cyclomatic complexity of this function is 123.
phobos/std/uni/package.d(6259:35)[warn]: Cyclomatic complexity of this function is 85.
phobos/std/process.d(885:13)[warn]: Cyclomatic complexity of this function is 51.
phobos/std/stdio.d(5457:16)[warn]: Cyclomatic complexity of this function is 96.
phobos/std/string.d(6071:6)[warn]: Cyclomatic complexity of this function is 67.
phobos/std/variant.d(261:22)[warn]: Cyclomatic complexity of this function is 54.
```

rikkimax:
> Looks like no one switched this project over from Travis to GitHub Actions yet.

Indeed, only a couple of days ago I deactivated the branch protection for travis for this repository and dfmt's.

src/dscanner/analysis/config.d
CyberShadow: One way to think of this is that the first class is useful to everyone, but the second class is useful only for people working in teams. If you are working on a personal or one-shot throwaway project, this class of warnings would be completely useless. So I'm actually wondering if it should perhaps be disabled by default.

rikkimax: > One way to think of this is that the first class is useful to everyone, but the second class is useful only for people working in teams. If you are working on a personal or one-shot throwaway project, this class of warnings would be completely useless. So I'm actually wondering if it should perhaps be disabled by default.

Even for a one-off, I would prefer to have that warning exist even if it's quite high just to ensure I don't start having bad practices.

CyberShadow:
> introduced phobos warnings:

Thanks! With links:

- https://github.com/dlang/phobos/blob/master/std/algorithm/searching.d#L1899 - 61
- https://github.com/dlang/phobos/blob/master/std/datetime/systime.d#L10305 - 59
- https://github.com/dlang/phobos/blob/master/std/experimental/logger/core.d#L2684 - 109
- https://github.com/dlang/phobos/blob/master/std/experimental/checkedint.d#L2543 - 69
- https://github.com/dlang/phobos/blob/master/std/format/internal/floats.d#L589 - 85
- https://github.com/dlang/phobos/blob/master/std/format/internal/write.d#L198 - 76
- https://github.com/dlang/phobos/blob/master/std/format/internal/write.d#L3375 - 56
- https://github.com/dlang/phobos/blob/master/std/internal/math/gammafunction.d#L764 - 72
- https://github.com/dlang/phobos/blob/master/std/internal/unicode_tables.d#L3778 - 60
- https://github.com/dlang/phobos/blob/master/std/math/exponential.d#L429 - 109
- https://github.com/dlang/phobos/blob/master/std/math/operations.d#L1492 - 55
- https://github.com/dlang/phobos/blob/master/std/net/isemail.d#L60 - 184
- https://github.com/dlang/phobos/blob/master/std/net/isemail.d#L1350 - 125
- https://github.com/dlang/phobos/blob/master/std/numeric.d#L1157 - 68
- https://github.com/dlang/phobos/blob/master/std/regex/internal/backtracking.d#L321 - 123
- https://github.com/dlang/phobos/blob/master/std/uni/package.d#L6259 - 85
- https://github.com/dlang/phobos/blob/master/std/process.d#L885 - 51
- https://github.com/dlang/phobos/blob/master/std/stdio.d#L5457 - 96
- https://github.com/dlang/phobos/blob/master/std/string.d#L6071 - 67
- https://github.com/dlang/phobos/blob/master/std/variant.d#L261 - 54

Seems reasonable to warn about these, though it's not clear what the action to resolve it is in each case.

I'm wondering if it would make sense to not reset the counter for nested functions. Right now, if you have

```d
void outer()
{
	// lots of code ...

	// Explanation for what this block does
	{
		// lots of code ...
	}

	// lots of code ...
}
```

you could silence the warning by doing

```d
void outer()
{
	// lots of code ...

	// Explanation for what this function does
	void doThing()
	{
		// lots of code ...
	}
	doThing();

	// lots of code ...
}
```

or perhaps even

```d
void outer()
{
	// lots of code ...

	// Explanation for what this block does
	(){
		// lots of code ...
	}();

	// lots of code ...
}
```

WebFreak001:
member functions should already work to silence the warnings, though adding a test for that would be good.

I explicitly made function literals (lambdas and stuff) not reset the counter, but instead even increase the counter by 1, as they in theory could be used like `something.ifThen({}, {})` (working around the if block) but also as actual practical thing they do add to the complexity of the function and we don't know if they are called conditionally or not - most of the time they are probably used conditionally or in a loop.

maxhaton:
How useful is this going to be in practice? Isn't this warning on the kind of thing that is obvious to anyone capable of "rectifying" the warning (which is usually not a simple refactor e.g. moving CFG nodes - does this keep track of what edges go where? - into places dscanner can't see as a way of shutting it up is not necessarily better than having things in one place).

WebFreak001:
this is just a simple implementation counting different control flow keywords. (e.g. see https://github.com/fzipp/gocyclo)

This can be useful in code bases to avoid too large / complex functions. Doing some code reviews at work I saw that code duplication has decreased and generally the functions have been split into multiple functions, doing specific things. I think it's an overall good improvement.

The default of 50 only really appears in huge functions (in D-Scanner there are just 3 occurrences) which shouldn't affect most code, and if it is something someone doesn't want to look at, you can disable it.

I think it's good to combine this score with something else, but for now this just adds the ability to calculate this one metric and an IDE can choose to make it emit for everything (0 threshold like in tests) and instead of showing it to the user with squiggly lines, use it with other metrics or to generate reports for example.

WebFreak001:
disabled the check by default as Max rightfully said that it may be less useful to general public, but rather to individuals seeking to have this feature.

Also disabled in dscanner for now as it emits a few warnings in D-Scanner due to very long repetitive functions.

PR #853, opened 2022-01-22
Add Github Actions setup to replace Travis

MoonlightSentinel:
The configured workflow tests D-Scanner with different host compilers
(DMD / LDC / GDC) on different platforms ( Windows / OSX / Linux).

The tests include
- `build.bat test` (windows) or `make test`
- `dub test` for version ranges of `libdparse` and `dsymbol`
   (currently disabled because of unresolvable dependency constraints)
- `make lint`
- phobos linting

---

[Last run](https://github.com/MoonlightSentinel/D-Scanner/actions/runs/1753237054)

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -1,0 +1,142 @@
+name: run-tests
+
+on:
+  pull_request:
+
+  push:
+    branches:
+      - master
+
+jobs:
+  main:
+    name: Run all tests
+
+    # Only run for the main repository - not forks
+    if: ${{ github.repository == 'dlang-community/D-Scanner' }}
+
+    # Run permutations of common os + host compilers
+    strategy:
+      fail-fast: false
+      matrix:
+        compiler: [
+          {
+            version: dmd-latest,
+            dmd: dmd
+          },
+          {
+            version: ldc-latest,
+            dmd: ldmd2
+          },
+          {
+            # Install dmd for the associated tools (dub/rdmd)
+            version: dmd-latest,
+            dmd: gdmd
+          }
+        ]
+        host: [
+          ubuntu-latest,
+          macos-latest,
+          windows-latest
+        ]
+        build: [
+          { type: make },
+          { type: dub, version: 'min libdparse' },
+          # Fail due to unresolvable dependencies
+          # { type: dub, version: 'max libdparse' },
+          # { type: dub, version: 'min dsymbol' },
+          # { type: dub, version: 'max dsymbol' },
+        ]
+
+        exclude:
+          # Restrict GDC to Ubuntu
+          - compiler:
+              dmd: gdmd
+            host: windows-latest
+          - compiler:
+              dmd: gdmd
+            host: macos-latest
+
+          #  Omit dub builds for GDC because dub rejects the old fronted revision
+          - compiler:
+              dmd: gdmd
+            build:
+              type: dub
+
+    runs-on: ${{ matrix.host }}
+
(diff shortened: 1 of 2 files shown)


WebFreak001:
dropped the tests with oldest and newest supported libdparse/dsymbol?

.github/workflows/default.yml
```diff
@@ -0,0 +1,146 @@
...
+            # echo "$FILE"
+            ../bin/dscanner -S --config=.dscanner.ini "$FILE"
+            EXIT=$?
+            case $EXIT in
+              0)
+                # Success
+                ;;
+
+              1)
+                echo "Warning[$FILE]: Exited with errors"
```
WebFreak001, line 133: why not exit 1? Seems reasonable to exit if an exception in dscanner is thrown.

.github/workflows/default.yml
```diff
@@ -0,0 +1,146 @@
...
+                ;;
+
+              137)
+                # Some files seem to require a lot of memory?
+                echo "Warning[$FILE]: Killed due to insufficient memory!"
```
WebFreak001, line 138: This seems like it would be a bug in dscanner, the github actions runners have 7 GB of RAM on Windows/Linux (14 GB of RAM on OSX) so that should never be reached through dscanner.

WebFreak001 (changes requested):

MoonlightSentinel:
> dropped the tests with oldest and newest supported libdparse/dsymbol?

Overlooked the additional permutations, will add them here.

.github/workflows/default.yml
```diff
@@ -0,0 +1,146 @@
...
+  push:
+    branches:
+      - master
+
+jobs:
+  main:
+    name: Run all tests
+
+    # Only run for the main repository - not forks
+    if: ${{ github.repository == 'dlang/dmd' }}
```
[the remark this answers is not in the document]
MoonlightSentinel: Note that this only disables the execution for plain branch pushes. This workflow will still be executed for PR's from forks

.github/workflows/default.yml
MoonlightSentinel: The errors and OOM occurs consistently on master. Initially didn't want to introduce a CI check that doesn't pass for the base branch but I'll change it s.t. the failure is reported.

.github/workflows/default.yml
MoonlightSentinel: See https://github.com/dlang-community/D-Scanner/pull/853#discussion_r792643324

.github/workflows/default.yml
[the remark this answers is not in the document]
WebFreak001: but why dlang/dmd?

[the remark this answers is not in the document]
MoonlightSentinel: Because I apparently had discarded the change to `dlang-community/D-Scanner`. Fixed

.github/workflows/default.yml
MoonlightSentinel: Turns out the failure was caused by using an outdated base branch. Sorry for the confusion.

The check will now fail eagerly for the first failure.

MoonlightSentinel:
Added the permutations. Also had to fix missing import paths/source files in `build.bat`.

WebFreak001 (approved):
LGTM

I noticed the here disabled libdparse failed dependency resolution cases, which is weird because it includes the min version, just as range. I think it's a bug in DUB.

PR #855, opened 2022-02-07
Modify makefile to use incremental compilation

edi33416:
Use incremental compilation to speed up development time

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,11 +1,11 @@
-.PHONY: all test
+.PHONY: all test clean
 
 DC ?= dmd
 GIT ?= git
 DMD := $(DC)
 GDC := gdc
 LDC := ldc2
-OBJ_DIR := obj
+
 LIB_SRC := \
 	$(shell find containers/src -name "*.d")\
 	$(shell find dsymbol/src -name "*.d")\
@@ -17,6 +17,13 @@ LIB_SRC := \
 	$(shell find stdx-allocator/source -name "*.d")
 PROJECT_SRC := $(shell find src/ -name "*.d")
 SRC := $(LIB_SRC) $(PROJECT_SRC)
+
+UT_OBJ_DIR = unittest-obj
+OBJ_DIR := obj
+OBJ = $(SRC:.d=.o)
+PROJECT_OBJ = $(PROJECT_SRC:.d=.o)
+LIB_OBJ = $(LIB_SRC:.d=.o)
+
 INCLUDE_PATHS = \
 	-Isrc \
 	-Iinifiled/source \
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -33,41 +40,78 @@ override DMD_FLAGS += $(DFLAGS)
...
+override LDC_FLAGS += -O5 -release -oq -d-version=StdLoggerDisableWarning -Jbin
+override GDC_FLAGS += -O3 -frelease -d-version=StdLoggerDisableWarning -Jbin
 SHELL:=/usr/bin/env bash
 
-all: dmdbuild
-ldc: ldcbuild
-gdc: gdcbuild
+DMD_BIN = bin/dmd/dscanner
+LDC_BIN = bin/ldc/dscanner
+GDC_BIN = bin/gdc/dscanner
```
WebFreak001, line 49: the executable should not be grouped by compiler, also previous tests depend on it being in bin/dscanner as well as this no longer matching dub builds.

WebFreak001 (changes requested):

makefile
```diff
@@ -33,41 +40,78 @@ override DMD_FLAGS += $(DFLAGS)
...
+UT_OBJ_BY_DMD = $(addprefix $(UT_OBJ_DIR)/dmd/, $(PROJECT_OBJ))
+
+OBJ_BY_LDC = $(addprefix $(OBJ_DIR)/ldc/, $(OBJ))
+UT_OBJ_BY_LDC = $(addprefix $(UT_OBJ_DIR)/ldc/, $(PROJECT_OBJ))
+
+OBJ_BY_GDC = $(addprefix $(OBJ_DIR)/gdc/, $(OBJ))
+UT_OBJ_BY_GDC = $(addprefix $(UT_OBJ_DIR)/gdc/, $(PROJECT_OBJ))
+
+$(OBJ_DIR)/dmd/%.o: %.d
+	@test -d $(dir $@) || mkdir -p $(dir $@)
```
MoonlightSentinel, line 63: Use a [order-only dependency](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html#Prerequisite-Types) for the directory

makefile
```diff
@@ -33,41 +40,78 @@ override DMD_FLAGS += $(DFLAGS)
...
+	${DC} ${GDC_TEST_FLAGS} ${VERSIONS} -unittest ${INCLUDE_PATHS} -c $< -o $@
+
+all: ${DMD_BIN}
+ldc: ${LDC_BIN}
+gdc: ${GDC_BIN}
 
-githash:
-	mkdir -p bin && ${GIT} describe --tags > bin/githash.txt
+${GITHASH}:
```
MoonlightSentinel, line 90: Keep the original alias:

```suggestion
githash: ${GITHASH}

${GITHASH}:
```

makefile
edi33416: The reason I grouped by compiler was to ensure that linked object files were all compiled by the same compiler.
I fear that otherwise (linking objects compiled by different compilers) might lead to UB.

makefile
edi33416: keeping the alias makes the of the objects relink everytime, regardless of changes

MoonlightSentinel: Only if the other targets depend on the alias instead of the file?

makefile
WebFreak001: you can group object files by compiler, but the resulting exe path must not change

edi33416:
I don't understand why the `gdmd` build is failing. The `.o` files are created (based on the build output) but `gdc` complains that the files don't exist. Any ideas?

RazvanN7 (approved):

edi33416:
@WebFreak001 @MoonlightSentinel is this good to go?

makefile
```diff
@@ -17,6 +17,25 @@ LIB_SRC := \
 	$(shell find stdx-allocator/source -name "*.d")
 PROJECT_SRC := $(shell find src/ -name "*.d")
 SRC := $(LIB_SRC) $(PROJECT_SRC)
+
+UT_OBJ_DIR = unittest-obj
```
WebFreak001, line 21: doesn't this clutter the project directory / add files that may accidentally be committed to git? (not gitignored)

I think it would be better if it was in obj/unittest or if the unittest-obj folder was added to gitignore.

RazvanN7:
ping @edi33416

makefile
edi33416: You're right. I've moved it inside the `obj` directory

WebFreak001 (approved):
LGTM, not a makefile user myself though

PR #857, opened 2022-02-13
Change rdmd to $DC -run

### LINKED ISSUE #821: Replace rdmd with dub?
As DMD x86_64 does not have a rdmd x86_64 executable, I see the error message `rdmd is not recognized as internal or external command` while building dscanner dub package. (I only have bin64 folder in PATH environment variable).

I wonder whether we can replace rdmd with a single file dub package.
https://github.com/dlang-community/D-Scanner/blob/2963358eb4a24064b0893493684d4075361297eb/dub.json#L27

Quickfix is of course adding `bin` folder to PATH or copying `rdmd` into folder `bin64`. (Also created a request to include x86_64 bit version of rdmd into DMD windows package).

WebFreak001:
Fix #821

LDC has the option -run, but GDC doesn't seem to have it yet.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -23,7 +23,7 @@
   "stringImportPaths" : [
     "bin"
   ],
-  "preGenerateCommands" : [
-    "rdmd \"$PACKAGE_DIR/dubhash.d\""
+  "preBuildCommands" : [
+    "$DC -run \"$PACKAGE_DIR/dubhash.d\""
   ]
 }


WebFreak001:
I'm not sure why CI says "Invalid source/import path: /home/runner/work/D-Scanner/D-Scanner/bin" after running the prebuild command, need to investigate what exactly happens here first

WebFreak001:
the issue is fixed with https://github.com/dlang/dub/pull/2217

For older DUB versions it's possible to run DUB with `env DC=dmd` (or other) to make it replace the D compiler.

The setup-dlang Action for GH Actions already sets the DC environment variable so the builds will work.

However until the DUB version is released (and we can bump the toolchain requirements), I will leave this unmerged.

rikkimax:
New dub version was released with PR.

PR #861, opened 2022-03-17
add test for ignored & templated unittests

WebFreak001:
in DMD it seems to be a syntax error to have unittests in functions or
other unittests, but in libdparse it's allowed, so this dscanner check
adds a warning/error for that.

In the same check it implements a warning for unittests which are used
in templates. However as unittests in templates could be there just for
documentation, no warning is issued if annotated with a doc comment.

Mixin templates are not regarded as templates in this check as they
might be used for automatically generating unittests outside templates.

### DIFF
--- a/src/dscanner/analysis/config.d
+++ b/src/dscanner/analysis/config.d
@@ -215,6 +215,9 @@ struct StaticAnalysisConfig
 	@INI("Maximum cyclomatic complexity after which to issue warnings")
 	int max_cyclomatic_complexity = 50;
 
+	@INI("Check for ignored unittests in templates or functions")
+	string ignored_unittest = Check.enabled;
+
 	@INI("Module-specific filters")
 	ModuleFilters filters;
 }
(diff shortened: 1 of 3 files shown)


rikkimax:
> In the same check it implements a warning for unittests which are used
> in templates. However as unittests in templates could be there just for
> documentation, no warning is issued if annotated with a doc comment.

Unittests should run in the binary that initialized them. If this isn't happening that is a bug.

That should probably be removed.

Error for this causing CI failures.

> std/experimental/allocator/building_blocks/bitmapped_block.d
std/experimental/allocator/building_blocks/bitmapped_block.d(13[70](https://github.com/dlang-community/D-Scanner/runs/5586003765?check_suite_focus=true#step:10:70):17)[warn]: unittest blocks in templates may not be run and can only serve for documentation purpose. Document with a ddoc-comment or move unittest block out of template.
std/experimental/allocator/building_blocks/bitmapped_block.d(1645:17)[warn]: unittest blocks in templates may not be run and can only serve for documentation purpose. Document with a ddoc-comment or move unittest block out of template.
std/experimental/allocator/building_blocks/bitmapped_block.d(2176:13)[warn]: unittest blocks in templates may not be run and can only serve for documentation purpose. Document with a ddoc-comment or move unittest block out of template.

PR #866, opened 2022-04-29
Update libdparse to latest version to fix ldc failure

RazvanN7:
See: https://github.com/dlang-community/libdparse/pull/458


RazvanN7:
This still needs a new tag from libdparse and then I can update the dub.json file.

WebFreak001 (approved):

PR #870, opened 2022-05-17
[DON'T PULL] Empty commit to trigger build and check for linkage error

edi33416:


rikkimax:
I don't know what the goal is here but looks like a template emission bug.

Possibly fixed by just emitting all template instances.

> /usr/bin/ld: obj/gdmd/libdparse/src/dparse/trivia.o: in function `_D3std9algorithm9searching44__T10countUntilVAyaa6_61203d3d2062TAAyaTAyaZ10countUntilFNaNbNiNfAAyaAyaZl':
trivia.d:(.text._D3std9algorithm9searching44__T10countUntilVAyaa6_61203d3d2062TAAyaTAyaZ10countUntilFNaNbNiNfAAyaAyaZl[_D3std9algorithm9searching44__T10countUntilVAyaa6_61203d3d2062TAAyaTAyaZ10countUntilFNaNbNiNfAAyaAyaZl]+0x1d): undefined reference to `_D3std9algorithm9searching139__T10countUntilS114_D3std9algorithm9searching44__T10countUntilVAyaa6_61203d3d2062TAAyaTAyaZ10countUntilFAAyaAyaZ5pred2MFNaNbNiNfAyaZbTAAyaZ10countUntilMFNaNbNiNfAAyaZl'

edi33416:
> I don't know what the goal is here but looks like a template emission bug.

We noticed that our [fork](https://github.com/Dlang-UPB/D-scanner) is now failing to build with gdc and I wanted to check if it reproduces in upstream as well.

> Possibly fixed by just emitting all template instances.

How can you do that with `gdc`? Does it have a switch similar to dmd's `-allinst`?

rikkimax:
I found this on the Wiki:

``-femit-templates``

https://wiki.dlang.org/Using_GDC

edi33416:
> I found this on the Wiki:
> 
> `-femit-templates`
> 
> https://wiki.dlang.org/Using_GDC

Apparently it's `-fall-instantiations` now.
Found it through `gdc --help=d`, in case someone else needs it

edi33416:
It seems that emitting all the template instances doesn't solve the issue

edi33416:
@WebFreak001 it seems that the `gdmd` build is still failing. Any ideas why?

WebFreak001:
hm the gdmd test seems a little weird, it installs the latest dmd and then uses gdc.

An issue here could be that the std.experimental.allocator library is old and potentially buggy.

WebFreak001:
it seems allinstantiations works on gdc-12, so closing this because of #877

Issue #880, opened 2022-12-28
package modifiers wrongly marked as equal


brianush1:
```d
module foo.bar.baz;

class Foo {
package:

	package(foo) this() {
		
	}

}

```

gives `same visibility attribute used as defined on line 4. DScanner(dscanner.suspicious.redundant_attributes)` on line 6

PR #881, opened 2022-12-28
fix #791

### LINKED ISSUE #791: Static `@property` function wrongly marked as needing `const`, `inout`, or `immutable`
```d
private class Foo {
	static int bar() @property {
		return 2;
	}
}
```
D-scanner suggests putting a `const`, `inout`, or `immutable` attribute on `bar`; however, doing this would result in a compilation error.

brianush1:


WebFreak001 (changes requested):
this doesn't seem to change the `package` / `package(xyz)` behavior referenced in #880, but only disables this check inside aggregates. (which seems wrong)

Also this is missing a unittest to check the code from the referenced issue.

brianush1:
different issue, this PR is for an issue I opened in 2020

WebFreak001:
ah sorry, missed that. This is still missing a test for the code from that issue though.

brianush1:
just added tests, not sure if github notifies for that so commenting this just in case

src/dscanner/analysis/function_attributes.d
```diff
@@ -147,3 +148,39 @@ private:
...
+			int barConst() const @property { return 0; }
+		}
+
+		interface InterfaceName {
+			int bar() @property; // [warn]: Zero-parameter '@property' function should be marked 'const', 'inout', or 'immutable'.
+			int barConst() const @property;
+
+			abstract int method(); // [warn]: 'abstract' attribute is redundant in interface declarations
+		}
+	}c, sac);
```
WebFreak001, line 183: can you also test static methods? (as mentioned in #791)

brianush1: done

WebFreak001 (approved):

PR #887, opened 2023-01-31
Fix noreturn discard warning

### LINKED ISSUE #886: Discarded return value warning does not ignore noreturn if defined locally
If I define `noreturn` as `typeof(*null)` in the current file or in an included `object.d` file (using my own custom runtime), then dscanner starts generating warnings that the `noreturn` value is being discarded.

```d
alias noreturn = typeof(*null);

noreturn foo() {
    while (1) {}
}

noreturn bar() {
    foo(); // warning: function value is discarded
}
```

If `noreturn` is not declared locally then it works as expected.

zyedidia:
Fixes #886

### DIFF
--- a/src/dscanner/analysis/unused_result.d
+++ b/src/dscanner/analysis/unused_result.d
@@ -37,12 +37,14 @@ private:
 public:
 
     const(DSymbol)* void_;
+    const(DSymbol)* noreturn_;
 
     ///
     this(string fileName, const(Scope)* sc, bool skipTests = false)
     {
         super(fileName, sc, skipTests);
         void_ = sc.getSymbolsByName(internString("void"))[0];
+        noreturn_ = sc.getSymbolsByName(internString("noreturn"))[0];
     }
 
     override void visit(const(ExpressionStatement) decl)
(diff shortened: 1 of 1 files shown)


WebFreak001 (approved):

WebFreak001:
oh, can you add a test?

src/dscanner/analysis/unused_result.d
```diff
@@ -37,12 +37,14 @@ private:
...
 
     const(DSymbol)* void_;
+    const(DSymbol)* noreturn_;
 
     ///
     this(string fileName, const(Scope)* sc, bool skipTests = false)
     {
         super(fileName, sc, skipTests);
         void_ = sc.getSymbolsByName(internString("void"))[0];
+        noreturn_ = sc.getSymbolsByName(internString("noreturn"))[0];
```
WebFreak001, line 47: CI fails here because noreturn might not exist it seems

WebFreak001 (changes requested):

src/dscanner/analysis/unused_result.d
zyedidia: Thanks, I have added a check to make sure it exists first (and added a unit test).

WebFreak001 (approved):
while a proper lookup mechanism (e.g. check if it's `typeof(*null)`) would be better, I think this is a reasonable implementation and doing it differently would probably require changes in dsymbol.

PR #888, opened 2023-02-01
Include aliasName in resolveSymbol

### LINKED ISSUE #885: Discarded return value warning does not work for aliased types
If a type is aliased, using a static method on the aliased type does not generate a warning if the return value is discarded:

```d
struct Foo {
    static bool val() {
        return false;
    }
}

alias X = Foo;

void main() {
    Foo.val(); // warning
    X.val();   // no warning
}
```

zyedidia:
Fixes #885.


WebFreak001 (approved):
ah I thought we would need some changes in dsymbol to resolve this properly, but if this is all that it takes that's good

PR #894, opened 2023-02-25
don't use deprecated properties

Moth-Tolias:
this silences some of the deprecation warnings emitted when building.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -12,7 +12,7 @@
     "StdLoggerDisableWarning"
   ],
   "dependencies": {
-    "libdparse": ">=0.22.0 <0.23.0",
+    "libdparse": ">=0.21.0 <0.23.0",
     "dcd:dsymbol": ">=0.14.0 <0.16.0",
     "inifiled": "~>1.3.1",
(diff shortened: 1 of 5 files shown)


WebFreak001 (changes requested):
you would need to bump the minimum libdparse version for this change. (see CI)

Moth-Tolias:
should work now?

Moth-Tolias:
oh dear, that doesn't look right...

Moth-Tolias:
now tested to work with `dub upgrade`.

WebFreak001:
you forgot to set the exact version to use when compiling D-Scanner as executable in dub.selections.json

the version range in dub.json is for library users to support multiple libdparse versions

WebFreak001:
it seems that the new dsymbol versions doesn't work with old libdparse versions, you probably have to bump the minimum libdparse version some more.

Moth-Tolias:
if it doesn't work this time i will become jokerfied

WebFreak001 (approved):

PR #901, opened 2023-05-09
upgrade libdparse to 0.23.0

WebFreak001:
- XML printer still prints `<arguments>`, now each argument is inside a `<argument>` which may have an attribute `named="token"`
- mismatched argument names are now always emitted, even if there are non-identifier arguments. It's possible to suppress them now by using named arguments

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -11,8 +11,8 @@
     "built_with_dub"
   ],
   "dependencies": {
-    "libdparse": ">=0.21.1 <0.23.0",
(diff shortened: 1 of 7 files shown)


WebFreak001:
since either D-Scanner needs to be merged with breaking CI or Phobos needs to be merged with breaking CI (due to needed syntax changes that the old D-Scanner doesn't understand), I think it makes more sense to merge D-Scanner with broken CI and will just merge it here.

PR #925, opened 2023-07-28
feat(highlight): support multiple themes

### LINKED ISSUE #924: Support additional themes for HTML highlighting
Currently, the HTML highlighting supports only Solarized Dark, and is hardcoded here: https://github.com/dlang-community/D-Scanner/blob/a958f9ac7b1654cde0e15689fc9c60623b55e6ce/src/dscanner/highlighter.d#L24

We can provide a `--theme` flag that allows users to pick from a small predefined set of themes. If anything, it provides an alternative to those who dislike Solarized Dark :P.

Would a PR be accepted for this?

snprajwal:
Fixes #924 

Implementing this led me to discover a bug, although I don't know if it has already been reported - initialising AAs with inlined structs will not parse:
```
immutable Theme[string] themes = [ "gruvbox": { "#000000", "#ffffff", ... } ];
```
This errors out during lexing, with a pretty long list of errors.

### DIFF
--- a/src/dscanner/highlighter.d
+++ b/src/dscanner/highlighter.d
@@ -10,8 +10,10 @@ import std.array;
 import dparse.lexer;
 
 // http://ethanschoonover.com/solarized
-void highlight(R)(ref R tokens, string fileName)
+void highlight(R)(ref R tokens, string fileName, string themeName)
 {
+	immutable(Theme)* theme = getTheme(themeName);
+
 	stdout.writeln(q"[
 <!DOCTYPE html>
 <html>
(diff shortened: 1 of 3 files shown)


src/dscanner/highlighter.d
```diff
@@ -76,3 +80,48 @@ void writeSpan(string cssClass, string value)
...
+	string num;
+	string str;
+	string op;
+	string type;
+	string cons;
+}
+
+immutable(Theme)* getTheme(string themeName)
+{
+	static immutable Theme solarized = {
```
WebFreak001, line 99: please don't use struct initializers like this, either use them to name your arguments (only needed when supporting old compilers) or use the auto generated struct ctor, which also can be used with named arguments (on newer compilers)

This syntax has been subject to deprecation before and I think once named arguments are widely supported it will quickly disappear

WebFreak001 (approved):

Issue #932, opened 2023-09-24
Respect `excludedSourceFiles` when scanning sources


chances:
My project interops with C and GLFW writes a lot of intermediate files with the `*.d` extension.

D-Scanner ought ignore files included in Dub's `excludedSourceFiles`.

WebFreak001:
D-Scanner is completely unrelated to DUB and this would be way out of scope, either specify the list of files you want to scan manually (e.g. using a shell script) or you could make a PR if you want some `--exclude` CLI argument that you can pass globs to ignore to.

WebFreak001:
I saw now in your PR that you probably want this with `dub lint`, in that case this is an issue on the DUB side that is invoking D-Scanner

chances:
See https://github.com/dlang/dub/issues/2700.

Thank you for reporting this issue, @WebFreak001.

WebFreak001:
sure that would be fine

PR #934, opened 2023-10-05
Add --exclude cli argument

ricardaxel:
Related #932

### DIFF
--- a/src/dscanner/main.d
+++ b/src/dscanner/main.d
@@ -199,20 +199,23 @@ else
 		}
 	}
 
-	if (excludePaths.length)
-	{
-		args  = expandArgs(args);
-
-		string[] newArgs = [args[0]];
-		foreach(arg; args[1 .. $])
+	auto expandedArgs = () {
+		auto expanded = expandArgs(args);
+		if (excludePaths.length)
 		{
-			if(!excludePaths.map!(p => arg.isSubpathOf(p))
-							.fold!((a, b) => a || b))
-				newArgs ~= arg;
-		}
+			string[] newArgs = [expanded[0]];
+			foreach(arg; args[1 .. $])
+			{
+				if(!excludePaths.map!(p => arg.isSubpathOf(p))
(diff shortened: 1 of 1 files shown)


src/dscanner/utils.d
```diff
@@ -128,6 +129,55 @@ string[] expandArgs(string[] args)
 	return rVal;
 }
 
+string absoluteNormalizedPath(in string path)
```
WebFreak001, line 132: make these private or package please, don't want to expose those as D-Scanner APIs.

src/dscanner/main.d
```diff
@@ -450,6 +466,9 @@ Options:
         modules. This option can be passed multiple times to specify multiple
         directories.
 
+	--exclude <file | directory>..., <file | directory>
```
WebFreak001, line 469: you used tabs in this string, use spaces for CLI output

src/dscanner/main.d
```diff
@@ -197,6 +199,21 @@ else
 		}
 	}
 
+	if (!excludePaths.empty)
```
WebFreak001, line 202:
```suggestion
	if (excludePaths.length)
```

since it's equivalent, but doesn't depend on importing std.range

src/dscanner/main.d
```diff
@@ -197,6 +199,21 @@ else
...
+	{
+		args  = expandArgs(args);
```
WebFreak001, line 204: the new args could be double-expanded, which I don't think would usually lead to problems, but is bad style still I think.

Can you add an `else` to this if block and call `args = expandArgs(args)` there, then strip the expandArgs from all further uses below?

WebFreak001 (changes requested):
thanks! see comments below for needed minor changes

src/dscanner/main.d
ricardaxel: This would mean that we always expands args, so we can even put this `args = expandArgs(args)` outside an if-else statement I think.

Issue #935, opened 2023-10-06
Skip lint on given line / function / class ..


ricardaxel:
Would you be open to add some mechanism to skip linting a given line / function / ... ?

For example through UDAs, following code wouldn't raise any warning.

```
@("NOLINT(useless_initializer)")
bool b = false;
```

I can work on a POC if you're ok with the concept

WebFreak001:
awesome! Yes this would be very much appreciated. Also it would be great if this worked both with `@("nolint(...)")` as well as `@nolint("...")` (for future compatibility or big projects that declare this)

We had the idea before that dub packages could provide files that are always implicitly imported by the compiler, such as `object` is already, which could provide exactly UDA definitions like this and help the compiler statically check for wrongly used attributes. However these UDA files will still need compiler flag support, which wasn't merged yet / didn't have a compelling use-case yet.

PR #936, opened 2023-10-08
Allow skipping checks with @("nolint(...)") and @nolint("...")

ricardaxel:
Related #935 

Here is the basic idea I had in mind to locally disable checks : one can attach a user defined attribute to any declaration (function declaration, variable declaration, ..). If this happens, the given check is disabled for the given declaration. 

Example : 
```
@("NOLINT(useless_initializer)")
int a = 0; // No warning emited
```

One could have done that with comment (like [clang-tidy](https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics) does) but I think UDAs are great since they are directly included in AST. They have a limitation though : according to [libdparse D grammar](http://libdparse.dlang.io/grammar.html#declaration), attributes can only be paired with 'declaration2'. So this would be a little hard to disable checks at other levels than declaration. For example, delete_check is applied on a deleteExpression.

Looking forward your reviews, I will try yo extend this proof of concept to all warnings in next few days if you're are okay with this direction.

### DIFF
--- a/src/dscanner/analysis/base.d
+++ b/src/dscanner/analysis/base.d
@@ -1,8 +1,8 @@
 module dscanner.analysis.base;
 
-import dscanner.analysis.nolint;
 import dparse.ast;
 import dparse.lexer : IdType, str, Token, tok;
+import dscanner.analysis.nolint;
 import dsymbol.scope_ : Scope;
 import std.array;
 import std.container;
@@ -413,8 +413,9 @@ public:
 	*/
 	override void visit(const(ModuleDeclaration) moduleDeclaration)
 	{
-		if(stopLinting(moduleDeclaration))
-			return;
+		auto currNoLint = NoLintFactory.fromModuleDeclaration(moduleDeclaration);
+		noLint.push(currNoLint);
+		scope(exit) noLint.pop(currNoLint);
 
 		moduleDeclaration.accept(this);
 	}
@@ -426,12 +427,11 @@ public:
 	*/
 	override void visit(const(Declaration) decl)
 	{
-		const msgDisabled = maybeDisableErrorMessage(decl);
+		auto currNoLint = NoLintFactory.fromDeclaration(decl);
+		noLint.push(currNoLint);
+		scope(exit) noLint.pop(currNoLint);
 
 		decl.accept(this);
-
-		if(msgDisabled)
-			reenableErrorMessage();
 	}
 
 	AutoFix.CodeReplacement[] resolveAutoFix(
@@ -452,7 +452,7 @@ protected:
 
 	bool inAggregate;
 	bool skipTests;
-	int errorMsgDisabled;
+	NoLint noLint;
 
 	template visitTemplate(T)
 	{
@@ -467,42 +467,42 @@ protected:
 	deprecated("Use the overload taking start and end locations or a Node instead")
 	void addErrorMessage(size_t line, size_t column, string key, string message)
 	{
-		if(!errorMsgEnabled())
+		if(noLint.containsCheck(this.getName()))
 			return;
 		_messages.insert(Message(fileName, line, column, key, message, getName()));
 	}
 
 	void addErrorMessage(const BaseNode node, string key, string message, AutoFix[] autofixes = null)
 	{
-		if(!errorMsgEnabled())
+		if(noLint.containsCheck(this.getName()))
 			return;
 		addErrorMessage(Message.Diagnostic.from(fileName, node, message), key, autofixes);
 	}
 
 	void addErrorMessage(const Token token, string key, string message, AutoFix[] autofixes = null)
 	{
-		if(!errorMsgEnabled())
+		if(noLint.containsCheck(this.getName()))
 			return;
 		addErrorMessage(Message.Diagnostic.from(fileName, token, message), key, autofixes);
 	}
 
 	void addErrorMessage(const Token[] tokens, string key, string message, AutoFix[] autofixes = null)
 	{
-		if(!errorMsgEnabled())
+		if(noLint.containsCheck(this.getName()))
 			return;
 		addErrorMessage(Message.Diagnostic.from(fileName, tokens, message), key, autofixes);
 	}
 
 	void addErrorMessage(size_t[2] index, size_t line, size_t[2] columns, string key, string message, AutoFix[] autofixes = null)
 	{
-		if(!errorMsgEnabled())
+		if(noLint.containsCheck(this.getName()))
 			return;
 		addErrorMessage(index, [line, line], columns, key, message, autofixes);
 	}
 
 	void addErrorMessage(size_t[2] index, size_t[2] lines, size_t[2] columns, string key, string message, AutoFix[] autofixes = null)
 	{
-		if(!errorMsgEnabled())
+		if(noLint.containsCheck(this.getName()))
 			return;
 		auto d = Message.Diagnostic.from(fileName, index, lines, columns, message);
 		_messages.insert(Message(d, key, getName(), autofixes));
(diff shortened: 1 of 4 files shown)


src/dscanner/analysis/base.d
```diff
@@ -437,45 +453,129 @@ protected:
...
 	void addErrorMessage(Message.Diagnostic diagnostic, Message.Diagnostic[] supplemental, string key, AutoFix[] autofixes = null)
 	{
+		if(!errorMsgEnabled())
+			return;
 		_messages.insert(Message(diagnostic, supplemental, key, getName(), autofixes));
 	}
 
+    void reenableErrorMessage()
+    in(this.errorMsgDisabled > 0)
+  	{
```
WebFreak001, line 513: you have mixed spaces and tabs everywhere in this file, try to stay consistent with the existing style (only tabs in this case)

src/dscanner/analysis/useless_initializer.d
```diff
@@ -361,6 +368,17 @@ public:
 		NotKnown nk = NotKnown.init;
 	}, sac);
 
+	// passes
+	assertAnalyzerWarnings(q{
+    	  	@("NOLINT(useless_initializer)")
+			int a = 0;
```
WebFreak001, line 374: check that an error is actually emitted afterwards as well within the same test string

WebFreak001 (changes requested):
can you also add a check to the `ModuleDeclaration` that checks for the UDAs and then just returns or continues with recursion? Also people might use whitespace within `NOLINT()` and maybe we just want to lowercase it instead of uppercasing, since non-string UDA equivalents would be all lower-case or camel-case

src/dscanner/analysis/base.d
ricardaxel: Do you use any formatter ? I can't find any config, this would be easier to maintain style conventions.

ricardaxel:
> can you also add a check to the ModuleDeclaration that checks for the UDAs and then just returns or continues with recursion?

Not sure to understand well. Do you mean something like that ?

```
override void visit(const ModuleDeclaration mod)
{
    if(hasNoLintUda(mod))
       return;
    else
      accept(this)
}
```

ricardaxel:
Soo I think this is a first mergeable version.

I think I'll keep other improvements for futures PRs, especially : 

- adding support and unittest for all checks one by one
- adding support for @("nolint"), @("nolint(useless_*)"),  ...

Just a question : what do you expect when a malformed UDA is given ? For example `@nolint("useles_initializer")` (it contains a typo). I guess raising an error is pretty extrem, and dump a warning could be confusing with real D-Scanner warning, no ?  For now, such cases are silently ignored

WebFreak001:
we could probably add a dscanner checker to check for misconfigured dscanner UDAs later

src/dscanner/analysis/nolint.d
```diff
@@ -0,0 +1,202 @@
...
+
+		return noLint.nullable;
+
+	}
+
+	// Transform a string with form "nolint(abc, efg)"
+	// into a NoLint struct
+	static Nullable!NoLint fromString(in string str)
+	{
+		auto re = regex(`\w+`, "g");
```
WebFreak001, line 162:
```suggestion
		auto re = regex(`[\w_.]+`, "g");
```

checks can contain underlines and also we probably want to filter by error key, not by linter type, since some linters emit different kinds of errors, where you might only want to exclude some. Also since it's split by `.` it would be quite easy to implement wildcards by simply using `startsWith`

Users would also be able to `@nolint("dscanner")` and since all keys start with `dscanner.`, we also leave open this UDA / string pattern for other linters in the ecosystem and can verify that dscanner keys are correctly written.

src/dscanner/analysis/useless_initializer.d
```diff
@@ -361,6 +364,33 @@ public:
 		NotKnown nk = NotKnown.init;
 	}, sac);
 
+	// passes
+	assertAnalyzerWarnings(q{
+			@("nolint(useless_initializer)")
+			int a = 0;
+		    ubyte a = 0x0;      /+
+		              ^^^ [warn]: X +/
```
WebFreak001, line 372:
```suggestion
			int a = 0; /+
			        ^ [warn]: X +/
```

try not changing here too much / being too smart, otherwise others reading this code might think this error comes from using 0x0 instead of 0 or only applies to ubyte and not int.

src/dscanner/analysis/useless_initializer.d
```diff
@@ -361,6 +364,33 @@ public:
...
+			int f() {
+				int[] a = null;
+			}
+
+			struct nolint { string s; }
+
+			@nolint("useless_initializer")
+			int a = 0;
+		    ubyte a = 0x0;      /+
+		              ^^^ [warn]: X +/
```
WebFreak001, line 384:
```suggestion
			int a = 0; /+
			        ^ [warn]: X +/
```

ditto

src/dscanner/analysis/style.d
```diff
@@ -232,5 +235,10 @@ unittest
 			     ^^^^^^^^^^^^^^ [warn]: Function name 'WinButWithBody' does not match style guidelines. +/
 	}c, sac);
 
+	assertAnalyzerWarnings(q{
+		@("nolint(style_check)")
+		module AMODULE;
```
WebFreak001, line 240:
```suggestion
		module mymod;
		int BAD_NAME;
```

ah the idea was here that the nolint would apply to the entire file, which is also what it currently does with the implementation. This test kind of implies that it would only apply to the module declaration itself and not the other code below though.

I think it's desirable to have the things apply to the whole file, so can you change the test to better reflect that? e.g. create a variable that has broken style name, but keep the module name valid.

For single instance checks like this we will probably want to apply the `nolint(...)` parsing to comments on the same line of the error as well, but that'd be for a different PR.

WebFreak001 (changes requested):
we need to make the nolint check check for error message IDs (in most linters this is the `enum KEY` member)

If you want to keep the optimization that it doesn't traverse into code, in the base analyzer class you probably want to add a `abstract string[] getAvailableErrorKeys();` to be able to check all of them where you currently only check for the check name, however there would need to be another check in the actual `addErrorMessage` code that will check before emitting the warning for real.

ricardaxel:
> we need to make the nolint check check for error message IDs

I totally missed that one analyzer could handle multiple checks. I made some refacto to handle this. 
The optimization to avoid the traversal is gone for now, and will probably be delayed for a future PR. 

> ah the idea was here that the nolint would apply to the entire file

I also completely misunderstood that, this is fixed :)

Hope I didn't miss anything else

WebFreak001 (approved):

AndrejMitrovic:
Great to see this feature in and thanks so much for working on this!

But I can't seem to silence this error for example:

```D
// test.d
@("nolint(dscanner.exception_check)")
void main() {
    try {
    } catch (Error) {
    }
}
```

```
$ dscanner lint ./test.d

.\test.d(4:14): Warning: Catching Error or Throwable is almost always a bad idea. (exception_check)
    } catch (Error) {
             ^^^^^
```

Using `master` of `dscanner`.

Also it would be nice to have this feature documented in `--help`, I only found out about it by accident.

And does https://github.com/dlang-community/D-Scanner/issues/935 need closing?

ricardaxel:
> Great to see this feature in and thanks so much for working on this!
> 
> But I can't seem to silence this error for example:
> 
> ```d
> // test.d
> @("nolint(dscanner.exception_check)")
> void main() {
>     try {
>     } catch (Error) {
>     }
> }
> ```

To silence an error, one has to use its ID (and not its name !), which is in this case  `dscanner.suspicious.catch_em_all`, as we can see [here](https://github.com/dlang-community/D-Scanner/blob/9076f7bab3845219b499de1b1e72d87d7f8d406b/src/dscanner/analysis/pokemon.d#L29). This is pretty confusing since the name and the key ID can be different, and this is the name which is used in the dscanner.ini configuration file...

> Also it would be nice to have this feature documented in `--help`, I only found out about it by accident.
> 
> And does #935 need closing?

The 'nolint' feature is pretty experimental, it has not been widely tested and would require some more polish (see previous messages of this PR + PR #941. I personally think that one should also work on adding coherence between names and IDs. So there is still some work to do). I think this is the reason for the lack of documentation.

PR #944, opened 2023-12-31
Update obsolete url

SixthDot:
in an old comment

### DIFF
--- a/src/dscanner/utils.d
+++ b/src/dscanner/utils.d
@@ -183,7 +183,7 @@ unittest
  * in other languages. In the chain, any $(D null) member that is a class instance
  * or that returns one, has for effect to shortcut the complete evaluation.
  *
- * This function is copied from https://github.com/BBasile/iz to avoid a new submodule.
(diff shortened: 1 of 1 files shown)


src/dscanner/utils.d
```diff
@@ -183,7 +183,7 @@ unittest
  * in other languages. In the chain, any $(D null) member that is a class instance
  * or that returns one, has for effect to shortcut the complete evaluation.
  *
- * This function is copied from https://github.com/BBasile/iz to avoid a new submodule.
+ * This function is copied from https://gitlab.com/basile.b/iz/-/blob/master/import/iz/sugar.d?ref_type=heads#L1543 to avoid a new submodule.
```
PetarKirov, line 186: Let's use a specific git commit:

```suggestion
 * This function is copied from
 * https://gitlab.com/basile.b/iz/-/blob/18f5c1e78a89edae9f7bd9c2d8e7e0c152f56696/import/iz/sugar.d#L1543
 * to avoid adding additional dependencies.
```

PetarKirov (approved):
Thanks!

PR #964, opened 2024-12-13
Replace libdparse with dmd as a library in static code analysis

Vladiwostok:

PR #971, opened 2025-08-31
Replace libdparse with dmd as a library in static code analysis

Albert24GG:
This is just a follow-up of #964 to hopefully get it merged.


Albert24GG:
@rikkimax Any idea why the pipeline fails?

WebFreak001:
why are all the error ranges being removed and replaced with just a line-based test? That will reduce usability of auto fixes and degrade IDE integration quite a lot. I think having proper start and end locations of warnings is pretty essential for D-Scanner and changing the expected behavior / changing the tests there is kinda a breaking change.

rikkimax:
> @rikkimax Any idea why the pipeline fails?

A quick look over the missing symbols error, shows that you're compiling dmd's glue code.

You want the ``:frontend`` sub package, rather than dmd as a whole: https://github.com/dlang/dmd/blob/master/dub.sdl#L84

PR #975, opened 2026-02-01
[style.alias_syntax] Add `AutoFix` replacement

ntrel:
Using `alias target ident;` syntax will be an error in the 2024 edition: https://github.com/dlang/dmd/pull/22244. (Perhaps `legacy.alias_syntax` might be a better setting key than `style.alias_syntax`, though maybe best not to change it if tools depend on the current name.)

This should handle all cases of old [alias syntax](https://dlang.org/spec/declaration.html#alias):
```d
    alias StorageClassesopt BasicType TypeSuffixesopt Identifiers ;
    alias StorageClassesopt BasicType FuncDeclarator ;
```
There is [a PR to update `druntime/src`](https://github.com/dlang/dmd/pull/22490). Although that was made using an updated `dfix`, I've checked the diff from this dscanner change and the only differences are on 2 lines where the trailing whitespace before `;` is different. I don't think they matter much.

Also extend `assertAnalyzerWarnings` tests.

I also made a `tokenStr` function - I'm not sure if a similar equivalent exists somewhere, it seems a useful general function.

---
Enhancement: When the target is a function pointer/delegate and there are multiple identifiers, it can generate long lines even when the original line was not too long. See the `LPCCHOOKPROC` part of the description [here](https://github.com/dlang/dmd/pull/22490). That mentions either using newlines before each identifier or changing the target for the remaining identifiers to the first identifier. Either of those should be straightforward to implement. The same can happen with other complicated targets. There could be a heuristic to enable different formatting when the target string length is more than e.g. 20.

### DIFF
--- a/src/dscanner/analysis/alias_syntax_check.d
+++ b/src/dscanner/analysis/alias_syntax_check.d
@@ -25,13 +25,64 @@ final class AliasSyntaxCheck : BaseAnalyzer
 
 	override void visit(const AliasDeclaration ad)
 	{
+		// new alias syntax uses `ad.initializers` instead
 		if (ad.declaratorIdentifierList is null)
 			return;
 		assert(ad.declaratorIdentifierList.identifiers.length > 0,
 				"Identifier list length is zero, libdparse has a bug");
+
+		static string tokenStr(const Token[] tokens)
+		{
+			string r;
+			foreach (i, t; tokens)
+			{
+				if (i)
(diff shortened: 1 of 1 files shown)


ntrel:
I don't know why one of the tests is failing, maybe unrelated.

WebFreak001 (approved):
thanks, looks great and fully tested!

Issue #640, opened 2018-05-25
Warning for variable used in ctor but not in standard func


BBasile:
```d
struct Foo {int i;}
void foo(int){}

void main()
{
    size_t i1; // [warn]: Variable i1 is never modified and could have been declared const or immutable.
    new Foo(i1);

    size_t i2; // nothing here
    foo(i2);
}
```

The bug is that there should not be any warning for `i1`.

Issue #298, opened 2016-01-09
False positive with unmodified variable check and functions that have ref/in/out parameters


kingsleyh:
Hi

This issue was raised on the intellij plugin - which uses dscanner for code linting:

https://github.com/kingsleyh/DLanguage/issues/68

It looks like an issue in Dscanner

BBasile:
The real dup is https://github.com/dlang-community/D-Scanner/issues/366 but since it contains a clear test case i close this one.

BBasile:
This is actually not exactly the same case.

Moth-Tolias:
just ran into this. any eta on a fix?

PR #978, opened 2026-08-04
could_be_immutable: fix false positives for ref/out parameters and member access

### LINKED ISSUE #325: don't check for could be immutable if object methods are called
Consider this example here:

```
import std.range: iota;
auto d = iota(2);
d.popFront;
assert(d.front == 1);
```

```
foo.d(5:10)[warn]: Variable d is never modified and could have been declared const or immutable.
```

And `dmd` obviously complains when adding `immutable` or `const`.

```
foo.d(6): Error: mutable method std.range.iota!(int, int).iota.Result.popFront is not callable using a immutable object
```

### LINKED ISSUE #366: False positive for Unused Variable
Consider the following code

```
    int opApply(scope int delegate(ref dchar) dg)
    {
        int result = 0;
        while (s.length != 0)
        {
            dchar c = decode(s);
            result = dg(c);
            if (result != 0) break;
        }
        return result;
    }
```

dscanner gives this warning:

```
std/encoding.d(585:19)[warn]: Variable c is never modified and could have been declared const or immutable.
```

dscanner should not emit this because the function called requires a ref parameter.

cschlote:
Fixes #366, fixes #325 (also covers the case reported in #298).

The check suggests `const`/`immutable` declarations that do not compile in two situations.

### 1. Arguments bound to `ref`/`out` parameters (#298 / #366)

```d
bool foo()
{
    bool overflowed;
    significand = mulu(significand, pow, overflowed); // "could be immutable" — FALSE
    if (overflowed)
        return false;
}
```

D-Scanner cannot know `mulu`'s signature (see the discussion in #366: "this limitation is unlikely to fall"), so a by-value and a `ref`/`out` binding cannot be distinguished. Arguments of function calls and `new` expressions are therefore now treated as potential modifications regardless of their type (new `callArgument` counter disabling the value-type early return in `variableMightBeModified`). The #640 behavior (`new Foo(i1)` does not warn) is preserved.

### 2. Member access without a call (#325)

```d
import std.range : iota;
auto d = iota(2);
d.popFront;                        // paren-less call — previously flagged
assert(d.front == 1);
```

As noted in #325, the AST cannot tell whether `d.popFront` is a call, and the member might be a non-`const` method/property (range accessors) or return something that requires a mutable base — e.g. `Nullable.get` on a struct holding an array, where `immutable doc` fails with "contains pointers or references". The base of any member access (`a.b`, chained `a.b.c`) is now treated as potentially modified (new `markMemberAccessBase`).

### Preserved behavior / trade-off

- Bare reads of value types are still reported: `int i = 1; return i;`
- Index reads are still reported: `int i = 0; return arr[i];`
- Variables used *only* through member access are no longer reported (`Point p; return p.x;`) — member const-ness cannot be proven syntactically.

### Testing

- All cases above added as unittests in `unmodified.d`, plus regression tests for the preserved warnings.
- `make test` (54 modules) and `tests/it.sh` pass.
- Running the patched binary over this repository's own sources removes exactly 7 `could_be_immutable` warnings — all confirmed false positives, e.g. the `equalRange` result range in `unmodified.d` itself — and adds none.

### DIFF
--- a/src/dscanner/analysis/unmodified.d
+++ b/src/dscanner/analysis/unmodified.d
@@ -142,10 +142,42 @@ final class UnmodifiedFinder : BaseAnalyzer
 
 	mixin PartsMightModify!AsmPrimaryExp;
 	mixin PartsMightModify!IndexExpression;
-	mixin PartsMightModify!FunctionCallExpression;
-	mixin PartsMightModify!NewExpression;
 	mixin PartsMightModify!IdentifierOrTemplateChain;
 	mixin PartsMightModify!ReturnStatement;
+
+	override void visit(const FunctionCallExpression functionCallExpression)
+	{
+		interest++;
+		if (functionCallExpression.type !is null)
+			functionCallExpression.type.accept(this);
+		if (functionCallExpression.unaryExpression !is null)
+			functionCallExpression.unaryExpression.accept(this);
+		if (functionCallExpression.templateArguments !is null)
+			functionCallExpression.templateArguments.accept(this);
+		// Arguments may be bound to `ref`/`out` parameters, which modifies
+		// them regardless of their type.
+		callArgument++;
+		if (functionCallExpression.arguments !is null)
+			functionCallExpression.arguments.accept(this);
+		callArgument--;
+		interest--;
+	}
+
+	override void visit(const NewExpression newExpression)
+	{
+		interest++;
+		if (newExpression.newAnonClassExpression !is null)
+			newExpression.newAnonClassExpression.accept(this);
+		if (newExpression.type !is null)
+			newExpression.type.accept(this);
+		callArgument++;
+		if (newExpression.arguments !is null)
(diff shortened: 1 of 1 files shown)


src/dscanner/analysis/unmodified.d
```diff
@@ -387,6 +445,126 @@ bool isValueTypeSimple(const Type type) pure nothrow @nogc
...
+			return 1;
+		}
+
+		long outParam()
+		{
+			bool createdNow;
+			immutable id = produce(createdNow);
+			return id + (createdNow ? 1 : 0);
+		}
+	}, sac);
```
WebFreak001, line 479: please add a test showing that a regular function parameter won't suppress the lint (out and ref parameters are not that common, they shouldn't be assumed to always be there and break the const lint)

WebFreak001 (changes requested):

